PR #140, opened 2024-03-18
Feature/modify account holder session

german664:
## Summary

-Add options in storybook to modify the session function and use different account holder, roles and apiKeys
-Make storybook our default playground environment, now to use playground we have to use `npm run playground`


isuert (commented):
Super nice to have this flexibility from storybook 👌

PR #633, opened 2025-04-07
FIX: Collapsed height computation of expandable card

glad-adyen:
## Summary
This PR fixes irregularities with the collapsed height computation of expandable cards with `inFlow` prop set to `true`, that break the expected rendering behavior.

The irregularity was first observed in a situation where the card elements were in the DOM tree, but hidden using a CSS `display: none` style declaration, so that the initial computation of the card's collapsed height was `0px`.


ElifDogaliBal (approved):

isuert (approved):
🐛🔨

PR #635, opened 2025-04-14
Fix github workflows

german664:
Modify github workflows to address security warnings


.github/workflows/compressed-size.yml
```diff
@@ -6,9 +6,13 @@ on:
       - main
       - develop
       - group/*
-
+      -
+permissions: read-all
```
glad-adyen, line 10: I see we already have permissions for the individual jobs, why do we still need this? Following the rule of least privileges, I'd suggest we skip this if not necessary.

german664: Mainly to avoid accidentally inheriting overly permissive defaults and enforce a secure default across all jobs. It could happen that we expand the workflow in the future and we might miss the permissions block. But I don't have a strong opinion, so I could remove it if you feel like it's a better option

glad-adyen: Yea — in general I would vote for removing them for now. And if we realize that we need more permissions for a workflow in the future, we can provision the permissions as needed.

glad-adyen: Let's do the same as needed for the other workflow files.

german664: ✅

.github/workflows/compressed-size.yml
```diff
@@ -6,9 +6,13 @@ on:
       - main
       - develop
       - group/*
+      -
```
ElifDogaliBal, line 9: Is this line needed?

german664: ✅

ElifDogaliBal (approved):

PR #644, opened 2025-04-16
Add missing workflows permissions

german664:


.github/workflows/upload-translations.yml
glad-adyen, line 1: Just curious, what about the `download-translations` workflow? Shouldn't we also have permissions configured for the jobs?

german664: It does! This PR is addressing the ones I missed in the previous one

glad-adyen: Great 👍

glad-adyen (approved):

PR #655, opened 2025-04-22
Upgrade storybook and vitest dependencies

german664:
## Summary

Upgrade storybook and vitest related dependencies to solve security vulnerabilities.


.storybook/runStorybook.ts
```diff
@@ -1,11 +1,8 @@
 import { buildDevStandalone } from '@storybook/core-server';
-import packageJson from '../package.json';
-import { PackageJson } from '@storybook/types';
 
 const STORYBOOK_DIR = __dirname;
 
 buildDevStandalone({
     configDir: STORYBOOK_DIR,
-    packageJson: packageJson as PackageJson,
```
glad-adyen, line 9: Just curious as to why we are removing this.

src/components/internal/Calendar/components/CalendarGrid/CalendarGridDayOfWeek.tsx
```diff
@@ -41,7 +41,7 @@ const CalendarGridDayOfWeek = ({ prepare, flags, labels: { long: longLabel, shor
     };
 
     const renderProps = getGridDayOfWeekRenderProps(props, prepare);
-    const { children: _, className, ...extendedProps } = renderProps.props || (EMPTY_OBJECT as NonNullable<typeof renderProps.props>);
+    const { children: _, className, onCommand, ...extendedProps } = renderProps.props || (EMPTY_OBJECT as NonNullable<typeof renderProps.props>);
```
glad-adyen, line 44: Why do we have this `onCommand` variable destructured here?

src/components/external/CapitalOverview/components/AccountDetails/AccountDetail.tsx
```diff
@@ -34,7 +34,7 @@ export const AccountDetail: FunctionalComponent<AccountDetailProps> = ({
             </dt>
             <dd className={contentClassName}>
                 {textToCopy ? (
-                    <CopyText buttonLabel={content} textToCopy={textToCopy} showCopyTextTooltip={false} type="Text" />
+                    <CopyText buttonLabel={content} textToCopy={textToCopy} showCopyTextTooltip={false} type={'Text' as const} />
```
glad-adyen, line 37: I notice the `as const` assertion being used with the `type` prop value for multiple instances of the `<CopyText />` component. Seems to me that the fault is with the `CopyText` component props type itself.

.storybook/runStorybook.ts
german664: It's no longer needed and was causing some unnecessary warnings in the console

src/components/external/CapitalOverview/components/AccountDetails/AccountDetail.tsx
german664: Not really. It was inferring the value of 'Text' as string, and the CopyText component expects  'Link' | 'Text' | 'Default'. So I think the component is right

src/components/internal/Calendar/components/CalendarGrid/CalendarGridDayOfWeek.tsx
german664: It was creating a type error and we were not using it so I decided to remove it from `extendedProps`.

glad-adyen: I am still not able to see the source of the `onCommand` property. Like where is it coming from? 🤔

glad-adyen (dismissed):

ElifDogaliBal (dismissed):

ElifDogaliBal (approved):

glad-adyen (approved):

PR #657, opened 2025-04-28
Dispute defend flow

ElifDogaliBal:
## Summary

- Add defend reason selection modal
- Get documents and render file upload modal
- Submit documents and open success or error modal
- Add new prop to Card Component to have the title and chevron horizontally aligned

**Fixed issue**:  [CXP-3042](https://youtrack.is.adyen.com/issue/CXP-3042/Implement-Defend-Dispute-flow)


src/components/external/DisputeManagement/components/AcceptDisputeFlow/AcceptDisputeFlow.tsx
```diff
@@ -37,8 +37,9 @@ export const AcceptDisputeFlow = ({
         options: {
             onSuccess: useCallback(() => {
                 onAcceptDispute?.();
+                clearStates();
                 setDisputeAccepted(true);
```
glad-adyen, line 41: What happens if `onAcceptDispute` throws an error here? Then the rest of the `onSuccess` logic doesn't run, and I don't think that is what we want in this case. I recommend doing this instead:

```suggestion
                clearStates();
                setDisputeAccepted(true);
                // always attempt to call this callback as the last step
                onAcceptDispute?.();
```

src/components/external/DisputeManagement/components/DisputeDetailsContainer/DisputeDetailsContainer.tsx
```diff
@@ -1,15 +1,18 @@
...
 
 export const DisputeDetailsContainer = ({
     disputeId,
+    onDefendDispute,
     onAcceptDispute,
     dataCustomization,
 }: {
     disputeId: string;
     onAcceptDispute?: () => void;
+    onDefendDispute?: () => void;
```
glad-adyen, line 15: I guess you will eventually get to it, but just for visibility — consider passing along this new `onDefendDispute` callback prop to the `<DefendDisputeFlow/>` component.

src/translations/en-US.json
glad-adyen, line 1: I guess you will eventually get to it. But just for visibility, please add translation instructions for the new translation keys.

src/components/external/DisputeManagement/components/DisputeDetails/DisputeDetails.tsx
```diff
@@ -1,22 +1,20 @@
...
 
-    const setDisputeCallback = useCallback((dispute: IDisputeDetail) => {
+    const setDisputeCallback = useCallback((dispute: IDisputeDetail | undefined) => {
         setDispute(dispute);
     }, []);
 
     return (
-        <DisputeFlowContext.Provider value={{ flowState, setFlowState, goBack, dispute: dispute, setDispute: setDisputeCallback }}>
+        <DisputeContextProvider dispute={dispute} setDispute={setDisputeCallback}>
             <DisputeDetailsContainer disputeId={props.id} onAcceptDispute={props.onAcceptDispute} dataCustomization={props.dataCustomization} />
```
glad-adyen, line 17: I guess you will eventually get to it. But just for visibility, consider adding `onDefendDispute` to the props here.

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
```diff
@@ -0,0 +1,74 @@
...
+    &__container {
+        display: flex;
+        flex-direction: column;
+        gap: style.token(spacer-070);
+        padding-top: style.token(spacer-070);
+        width: 100%;
+    }
+
+    &-document-requirements {
+        font-family: system-ui;
```
glad-adyen, line 13: I'd suggest we use `inherit` here for the `font-family` property.

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
```diff
@@ -0,0 +1,74 @@
...
+
+        &--list {
+            margin: 0;
+            padding-left: style.token(spacer-070);
+        }
+    }
+
+    &__title {
+        color: #000;
```
glad-adyen, line 22: Do we have a token for this color? If so, maybe we should consider using it.

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    &__error-icon {
+        color: style.token(color-background-critical-strong);
+        height: style.token(spacer-100);
+        width: style.token(spacer-100);
+
+        svg {
+            height: inherit;
+            width: inherit;
```
glad-adyen, line 33: I think what you need here is a style declaration to get the desired icon size.
```suggestion
            display: block;
            font-size: 16px;
```

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
```diff
@@ -0,0 +1,74 @@
...
+        }
```
glad-adyen, line 34: I think what you need for the `svg` element here is some style declaration(s) to get rid of the vertical alignment weirdness with `svg` elements.

To get the desired `svg` icon size, setting the correct font size on the `svg` parent element should suffice, since the `svg` element already has a `height` property value of `1em` (coming from the `<Icon />` component).

```suggestion
        color: style.token(color-background-critical-strong);
        font-size: style.token(spacer-100);

        svg {
            display: block;
        }
```

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
```diff
@@ -0,0 +1,74 @@
...
+
+    &__success-icon {
+        color: style.token(color-background-success-strong);
+        height: style.token(spacer-100);
+        width: style.token(spacer-100);
+
+        svg {
+            height: inherit;
+            width: inherit;
+        }
```
glad-adyen, line 72: Ditto: previous [comment](https://github.com/Adyen/adyen-platform-experience-web/pull/657/files#r2065997759)

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
```diff
@@ -0,0 +1,74 @@
...
+        padding: style.token(spacer-100) 0;
+        text-align: center;
+    }
+
+    &__error {
+        align-items: center;
+        display: flex;
+        flex-direction: column;
+        gap: style.token(spacer-070);
+        height: 218px;
```
glad-adyen, line 51: Just curious as to why we set an absolute `height` value here.

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
```diff
@@ -0,0 +1,74 @@
...
+        justify-content: center;
+    }
+
+    &__success {
+        align-items: center;
+        display: flex;
+        flex-direction: column;
+        gap: style.token(spacer-070);
+        height: 218px;
```
glad-adyen, line 60: Ditto: previous [comment](https://github.com/Adyen/adyen-platform-experience-web/pull/657/files#r2066006932)

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeReason.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const disputeId = dispute?.id;
+    const [isFetching, setIsFetching] = useState<boolean>(false);
+
+    //TODO: Fix the translations for defend reason
+    const defenseReasons: Readonly<{ id: string; name: string }[] | null> = useMemo(
+        () =>
+            Object.freeze(
+                allowedDefenseReasons?.map(reason => ({
+                    id: reason,
+                    name: i18n.has(`defendReason.${reason}`) ? i18n.get(`defendReason.${reason}` as TranslationKey) : reason,
```
glad-adyen, line 31: What do you think about renaming `defendReason` to `defenseReason` here to keep the consistency? Also, I remember we discussed some time ago about adding a sub-namespace for translation keys like this — something like: `disputes.defenseReason.{reason}`.

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeReason.tsx
```diff
@@ -0,0 +1,115 @@
...
+            onSuccess: useCallback(
+                ({ data }: { data: IApplicableDefenseDocument[] }) => {
+                    setIsFetching(false);
+                    setApplicableDocuments(data ?? null);
+                    if (data?.length > 0) setFlowState('uploadDefenseFilesView');
+                },
+                [setApplicableDocuments, setIsFetching, setFlowState]
+            ),
+        },
+    });
```
glad-adyen, line 56: I see we are using the `useMutation` hook for getting the applicable documents here. Have you considered using the `useFetch` hook? Since `useMutation` is usually considered as the option for `POST` requests.

src/components/external/DisputeManagement/components/DisputesData/DisputeData.tsx
```diff
@@ -119,12 +121,15 @@ export const DisputeData = ({
...
 
-    if (!dispute || isFetching) {
-        return <DataOverviewDetailsSkeleton skeletonRowNumber={5} />;
-    }
+    const onDefendClick = useCallback(() => {
+        dispute && setDispute(dispute);
+        if (dispute?.defensibility === 'defendable') {
+            setFlowState('defendReasonSelectionView');
+        }
+    }, [dispute, setDispute, setFlowState]);
```
glad-adyen, line 129: I don't think we need to `setDispute` explicitly here, since that already happens in the `onSuccess` callback of `useFetch` above. This also applies to the `onAcceptClick` callback.

Also, I'd suggest replacing the condition here with the `isDefendable` value. Similar can be done in the `onAcceptClick` callback.

```suggestion
    const onDefendClick = useCallback(() => {
        if (isDefendable) {
            setFlowState('defendReasonSelectionView');
        }
    }, [setFlowState]);
```

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFileUpload.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+    const defendDisputeMutation = useMutation({
+        queryFn: defendDispute,
+        options: {
+            onSuccess: useCallback(() => {
+                clearFiles();
+                onDefendSubmit('success');
+            }, []),
+            onError: useCallback(() => {
+                clearFiles();
```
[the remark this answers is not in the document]
ElifDogaliBal: At this point user tried to upload some files and we get an error from backend so, here we are clearing the previous files which the user tried to upload and get an error but it is not a blocker for re-upload, they can go back to upload modal and try uploading again

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeResponse.tsx
```diff
@@ -0,0 +1,47 @@
...
+                    </Button>
+                </div>
+            ) : (
+                <div className="adyen-pe-defend-dispute__error">
+                    <Icon name="cross-circle-fill" className="adyen-pe-defend-dispute__error-icon" />
+                    <Typography variant={TypographyVariant.TITLE}>{i18n.get('refundActionErrorTitle')}</Typography>
+                    <Button variant={ButtonVariant.SECONDARY} onClick={goBackToFileUploadView}>
+                        {i18n.get('disputes.goBack')}
+                    </Button>
+                </div>
```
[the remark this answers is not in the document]
ElifDogaliBal: We don't have an error cases yet but normally when we have [this success modal](https://www.figma.com/design/iJOHjJcQICUYirgX2r3Z9k/Disputes?node-id=19468-79352&t=UCGAY0ynSlY4gkAV-4) there is also error modal. I added error case consistent with what we have for [refund action error](https://www.figma.com/design/AhtlfOrOEU1dV65sf7htt2/Components---After-payment?node-id=13177-38602&t=cFnH8wXxJcAhTmyj-4) but this can be change based on design in the future

src/translations/en-US.json
```diff
@@ -112,20 +112,34 @@
...
     "dispute.evidence": "Evidence",
+    "dispute.evidenceSubmitted": "Evidence has been submitted",
     "dispute.merchantReference": "Merchant reference",
     "dispute.notDefended": "This chargeback wasn’t defended and was lost by default.",
     "dispute.paymentReference": "Payment reference",
     "dispute.reasonCode": "Reason Code",
+    "dispute.selectDefenseReasonDescription": "Select the reason you’re defending this dispute.",
+    "dispute.showDetails": "Show details",
+    "dispute.submit": "Submit",
+    "dispute.uploadDocumentChargebackFeeInformation": "Upload documents that support your dispute defense. Once submitted, no changes can be made.",
```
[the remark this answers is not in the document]
ElifDogaliBal: Good catch, changed :white_check_mark:

src/translations/en-US.instructions.json
```diff
@@ -71,6 +71,9 @@
     "dispute.account": "Label for a balance account",
     "dispute.contactSupportToDefendThisDispute": "Label to inform user that it is not possible to defend the dispute via component and they can contact support.",
     "dispute.defendDispute": "Label of an action button that opens a modal to upload some documents to defend the dispute",
+    "dispute.defendDocumentAcceptableFormats": "List item to inform user dispute defense documents should have only specified file formats",
+    "dispute.defendDocumentMustBeInEnglish": "List item to inform user dispute defense documents should be in English",
+    "dispute.defendDocumentRecommendedSize": "List item to inform user dispute defense documents should be A4 or US letter size",
```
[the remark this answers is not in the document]
ElifDogaliBal: :white_check_mark:

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFileUpload.tsx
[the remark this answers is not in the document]
isuert: I'm thinking maybe the user has uploaded several files and only one of them is faulty. In this case I'm fine with clearing the faulty file. Anyway clearing the other files that are already correct wouldn't make a good UX in my opinion. WDYT?

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
```diff
@@ -0,0 +1,74 @@
...
+        height: style.token(spacer-100);
+        width: style.token(spacer-100);
+
+        svg {
+            height: inherit;
+            width: inherit;
+        }
+    }
+
+    &__response {
```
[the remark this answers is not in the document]
ElifDogaliBal: I think we can leave this for now since we use the similar modal in accept dispute response and refund response too so, I think we can have a separate ticket to make this view an internal component and move the styles there in the scope of that ticket. For now there is a TODO for creating an internal component for this success/error modal. WDYT?

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
ElifDogaliBal: We don't have a token for this color and this font weight

glad-adyen: Then another option could be to communicate this discrepancies back to design. For the `color`, the primary color is close enough to pure black as used here (not really indistinguishable), and so can be used instead. I think the same is applicable to the `font-weight`.

src/components/external/DisputeManagement/components/DisputesData/DisputeData.tsx
```diff
@@ -114,28 +116,31 @@ export const DisputeData = ({
...
     const isDefendable = dispute?.dispute.defensibility === 'DEFENDABLE' && defendAuthorization;
     const isAcceptable = dispute?.dispute.defensibility === 'ACCEPTABLE' || dispute?.dispute.defensibility === 'DEFENDABLE';
 
+    const onAcceptClick = useCallback(() => {
+        if (isAcceptable) setFlowState('accept');
+    }, [isAcceptable, setDispute, setFlowState]);
+
+    const onDefendClick = useCallback(() => {
+        if (isDefendable) setFlowState('defendReasonSelectionView');
+    }, [isDefendable, setDispute, setFlowState]);
```
glad-adyen, line 129: `setDispute` no longer needed as a dependency here.

mocks/mock-data/disputes.ts
```diff
@@ -488,9 +488,12 @@ export const getApplicableDisputeDefenseDocuments = (dispute: (typeof DISPUTES)[
                 switch (disputeDefenseReason) {
                     case 'AirlineFlightProvided':
                         return [
+                            { documentTypeCode: 'DisclosureAtPointOfInteraction', requirementLevel: 'REQUIRED' } as const,
```
glad-adyen, line 491: Just curious as to where I can find this document requirement. Can't seem to find it anywhere?

src/components/external/DisputeManagement/components/DisputesData/DisputeData.tsx
```diff
@@ -119,23 +121,30 @@ export const DisputeData = ({
         setFlowState('accept');
     }, [dispute, setDispute, setFlowState]);
 
+    const onDefendClick = useCallback(() => {
+        dispute && setDispute(dispute);
+        if (dispute?.dispute.defensibility === 'DEFENDABLE') {
```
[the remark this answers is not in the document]
glad-adyen: @ElifDogaliBal Since you removed this condition from `onDefendClick` already, I think you should do the same for `onAcceptClick` (for the same reasons).

src/components/external/DisputeManagement/components/DisputesData/DisputeData.tsx
```diff
@@ -76,19 +76,20 @@ export const DisputeData = ({
...
 }) => {
     const { i18n } = useCoreContext();
-    const { setDispute, setFlowState } = useDisputeFlow();
+    const { dispute: storedDispute, setDispute, setFlowState } = useDisputeFlow();
 
     const { getDisputeDetail } = useConfigContext().endpoints;
 
     //TODO: Also check if /defend endpoint has been returned from setup call which relates to submit button action
     const defendAuthorization = isFunction(useConfigContext().endpoints.getApplicableDefenseDocuments);
+    const acceptAuthorization = isFunction(useConfigContext().endpoints.acceptDispute);
```
glad-adyen, line 85: I see you already have a destructuring statement for `useConfigContext().endpoints` in this component. Consider destructuring the endpoints for `acceptDispute` and `defendDispute` using the same statement to reduce verbosity.

src/components/external/DisputeManagement/components/DisputesData/DisputeData.tsx
```diff
@@ -114,28 +117,32 @@ export const DisputeData = ({
 
     const statusBoxOptions = useStatusBoxData(statusBoxProps);
 
+    const showContactSupport = !!dispute?.dispute.defensibility && ['ACCEPTABLE', 'DEFENDABLE_EXTERNALLY'].includes(dispute?.dispute.defensibility);
+    const isDefendable = !!dispute?.dispute.defensibility && dispute?.dispute.defensibility === 'DEFENDABLE' && defendAuthorization;
+    const isAcceptable =
+        !!dispute?.dispute.defensibility && ['ACCEPTABLE', 'DEFENDABLE'].includes(dispute?.dispute.defensibility) && acceptAuthorization;
```
glad-adyen, line 123: To reduce verbosity here, consider capturing `dispute?.dispute.defensibility` in a variable.
```suggestion
    const showContactSupport = !!defensibility && ['ACCEPTABLE', 'DEFENDABLE_EXTERNALLY'].includes(defensibility);
    const isDefendable = !!defensibility && defensibility === 'DEFENDABLE' && defendAuthorization;
    const isAcceptable = !!defensibility && ['ACCEPTABLE', 'DEFENDABLE'].includes(defensibility) && acceptAuthorization;
```

src/components/external/DisputeManagement/context/dispute/context.tsx
```diff
@@ -0,0 +1,128 @@
...
+    const [applicableDocuments, setApplicableDocuments] = useState<IDisputeDefenseDocument[] | null>([]);
+    const [defendDisputePayload, setDefendDisputePayload] = useState<any | null>(null);
+    const [defendResponse, setDefendResponse] = useState<'error' | 'success' | null>(null);
+
+    const clearFiles = useCallback(() => {
+        const formData = new FormData();
+        if (selectedDefenseReason) formData.set('defenseReason', selectedDefenseReason);
+        setSelectedDefenseReason(selectedDefenseReason);
+        setDefendDisputePayload(formData);
+    }, [selectedDefenseReason]);
```
glad-adyen, line 45: Since you have `selectedDefenseReason` as a dependency here rather than a callback argument, the `setSelectedDefenseReason` call is redundant and can be removed.

```suggestion
    const clearFiles = useCallback(() => {
        const formData = new FormData();
        if (selectedDefenseReason) formData.set('defenseReason', selectedDefenseReason);
        setDefendDisputePayload(formData);
    }, [selectedDefenseReason]);
```

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFileUpload.tsx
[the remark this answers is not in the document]
ElifDogaliBal: We talked offline but just to note here as well, checked with Marta and as @isuert suggested we will not clear the files after an error case

src/components/internal/ExpandableCard/ExpandableCard.tsx
```diff
@@ -23,7 +23,15 @@ import {
     CONTENT_EXPANDABLE_CLASS,
 } from './constants';
 
-const ExpandableCard = ({ renderHeader, children, filled, fullWidth, inFlow, ...listeners }: PropsWithChildren<ExpandableCardProps>) => {
+const ExpandableCard = ({
```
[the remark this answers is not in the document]
ElifDogaliBal: This new component is called expadable card in Bento and as a functionality it is same as expandable card only there is a visual change, so I removed structural changes and made the changes by the styles.

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
isuert: Adding to @glad-adyen's comment I'd suggest using specifically `color-label-primary` instead of #000 and `text-title-font-weight` instead of 700.

src/translations/en-US.json
```diff
@@ -114,6 +114,31 @@
     "default": "Default",
     "description": "Description",
     "dismiss": "Dismiss",
+    "dispute.account": "Account",
+    "dispute.contactSupportToDefendThisDispute": "%{contactSupport} to defend this dispute",
+    "dispute.continue": "Continue",
+    "dispute.defendDispute": "Defend dispute",
```
isuert, line 120: - I think we can use `disputes` namespace (in plural) for both list and details. Then we wouldn't need this key as it already exists
- "Defend dispute" has changed to "Defend chargeback". Same for accept.

src/components/external/DisputeManagement/components/DisputesData/DisputeData.tsx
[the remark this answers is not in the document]
ElifDogaliBal: done!

src/components/internal/ExpandableCard/ExpandableCard.tsx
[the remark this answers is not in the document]
ElifDogaliBal: I think that is a good point, I will change it and add it to Card component

src/components/external/DisputeManagement/components/DisputeDetails/DisputeDetails.tsx
```diff
@@ -1,22 +1,50 @@
...
-            <DisputeDetailsContainer disputeId={props.id} onAcceptDispute={props.onAcceptDispute} dataCustomization={props.dataCustomization} />
-        </DisputeFlowContext.Provider>
+    const isDefendFlow = useMemo(
+        () => ['defendReasonSelectionView', 'uploadDefenseFilesView', 'defenseSubmitResponseView'].includes(flowState),
+        [flowState]
     );
+
+    if (isDefendFlow) {
+        return <DefendDisputeFlow onDefendDispute={onDefendDispute} />;
+    }
```
glad-adyen, line 29: Instead of isolating the defend flow states here, why not just add these flow states to the `switch` statement below, to keep everything together?

```ts
switch (flowState) {
    case 'details':
        return (...);
    case 'accept':
        return (...);
    case 'defendReasonSelectionView':
    case 'defenseSubmitResponseView':
    case 'uploadDefenseFilesView':
        return <DefendDisputeFlow onDefendDispute={onDefendDispute} />;
    default:
        return null;
}
```

src/components/external/DisputeManagement/components/DisputeDetails/DisputeDetails.tsx
```diff
@@ -1,22 +1,50 @@
...
+
+    switch (flowState) {
+        case 'details':
+            return (
+                <>
+                    {flowState === 'details' && !hideTitle && (
```
glad-adyen, line 35: This `flowState === 'details'` check here is redundant, since the `flowState` value for this case is `'details'`.
```suggestion
                    {!hideTitle && (
```

src/components/external/DisputeManagement/components/DisputeDetails/DisputeDetails.tsx
```diff
@@ -1,22 +1,50 @@
...
+                        <div>
+                            <Typography variant={TypographyVariant.TITLE} medium>
+                                {i18n.get('dispute.management')}
+                            </Typography>
+                        </div>
```
glad-adyen, line 40: Is this wrapper `div` element necessary? You can use the `el={TypographyElement.DIV}` prop directly on the `Typography` component alongside some class names (if needed).

```suggestion
                        <Typography
                            el={TypographyElement.DIV}
                            variant={TypographyVariant.TITLE}
                            medium
                        >
                            {i18n.get('dispute.management')}
                        </Typography>
```

src/components/external/DisputeManagement/components/DisputeDetails/DisputeDetails.tsx
```diff
@@ -1,22 +1,50 @@
...
+                    )}
+                    <DisputeData disputeId={id} dataCustomization={dataCustomization} />
+                </>
+            );
+        case 'accept':
+            return <AcceptDisputeFlow disputeId={id} onBack={goBack} onAcceptDispute={onAcceptDispute} />;
```
glad-adyen, line 46: Since we are relying on context here, I don't think we need to pass `disputeId` and `onBack` as props to the `AcceptDisputeFlow` component anymore — they can be obtained from context within the component as needed. This will also keep the consistency with the `DefendDisputeFlow` component.

```suggestion
            return <AcceptDisputeFlow onAcceptDispute={onAcceptDispute} />;
```

src/components/external/DisputeManagement/components/DisputesData/DisputeData.tsx
glad-adyen, line 1: For rendering the dispute action buttons, consider changing the condition `isAcceptable || isDefendable` to something like `actionButtons.length > 0` — so that the buttons are only rendered if there is at least one, without repeating the checks that have already been done to determine which buttons are available.

For example:
```ts
{actionButtons.length > 0 && (
    <div className={DISPUTE_DATA_ACTION_BAR}>
        <ButtonActions actions={actionButtons} />
    </div>
)}
```

glad-adyen (changes requested):

src/components/internal/Card/types.ts
```diff
@@ -14,4 +15,7 @@ export interface CardProps {
     buttonAriaLabel?: string;
     classNameModifiers?: string[];
     testId?: string;
+    expandable?: boolean;
 }
+
+export type AriaRole = Pick<JSXInternal.HTMLAttributes<HTMLDivElement>, 'role'>;
```
glad-adyen, line 21: Consider changing this to `JSXInternal.HTMLAttributes<HTMLDivElement>['role']` to get the actual role type, rather than a record with a `role` property.

```suggestion
export type AriaRole = JSXInternal.HTMLAttributes<HTMLDivElement>['role'];
```

src/components/internal/Card/Card.tsx
```diff
@@ -31,6 +39,35 @@ const Card = ({
...
+        if (expandable) {
+            setShowContent(!showContent);
+        }
+    }, [expandable, showContent]);
+
+    const onKeyDown = useCallback(
+        (evt: KeyboardEvent) => {
+            if (evt.code === InteractionKeyCode.ENTER || evt.code === InteractionKeyCode.SPACE) {
+                toggleExpansion();
+            }
```
glad-adyen, line 55: I'd suggest you also prevent default action here.

```suggestion
            switch (evt.code) {
                case InteractionKeyCode.ENTER:
                case InteractionKeyCode.SPACE:
                    evt.preventDefault();
                    toggleExpansion();
                    return;
            }
```

src/components/internal/Card/Card.tsx
```diff
@@ -31,6 +39,35 @@ const Card = ({
...
 }: PropsWithChildren<CardProps>) => {
     const Tag = el || 'header';
+    const [showContent, setShowContent] = useState(false);
+    const cardId = useMemo(() => uuid(), []);
+
+    const toggleExpansion = useCallback(() => {
+        if (expandable) {
+            setShowContent(!showContent);
+        }
+    }, [expandable, showContent]);
```
glad-adyen, line 49: Instead of having `showContent` as a dependency here, you can switch to the computed version of the state setter.

```suggestion
    const toggleExpansion = useCallback(() => {
        if (expandable) setShowContent(showContent => !showContent);
    }, [expandable]);
```

src/components/internal/Card/Card.tsx
```diff
@@ -39,24 +76,34 @@ const Card = ({
                 classNameModifiers
             )}
             data-testid={testId}
+            {...cardContainerAttributes}
         >
             {(title || renderHeader) && (
-                <Tag className={CARD_HEADER}>
+                <Tag className={classNames(CARD_HEADER, { [`${CARD_HEADER}--expandable`]: expandable })}>
                     {(title || renderHeader) && (
```
glad-adyen, line 83: This duplicate `title || renderHeader` check not needed, since it has already been done upstream.

src/components/internal/Card/Card.tsx
```diff
@@ -39,24 +76,34 @@ const Card = ({
                 classNameModifiers
             )}
             data-testid={testId}
+            {...cardContainerAttributes}
         >
             {(title || renderHeader) && (
-                <Tag className={CARD_HEADER}>
+                <Tag className={classNames(CARD_HEADER, { [`${CARD_HEADER}--expandable`]: expandable })}>
```
glad-adyen, line 82: I'd recommend having one expandable state class at the top-most parent of all the elements (the `section` element), rather than having three different expandable state classes (for header, header content and body respectively).

src/components/internal/Card/Card.tsx
```diff
@@ -39,24 +76,34 @@ const Card = ({
...
                 </Tag>
             )}
-            <div
-                className={classNames(CARD_BODY, {
-                    [CARD_BODY_WITH_TITLE]: title || renderHeader,
-                })}
-            >
-                {children}
-            </div>
+            {(!expandable || (expandable && showContent)) && (
```
glad-adyen, line 96: Consider simplifying this check as follows:
```suggestion
            {(!expandable || showContent) && (
```

src/components/internal/Card/Card.tsx
```diff
@@ -39,24 +76,34 @@ const Card = ({
...
-                        <div className={CARD_HEADER_CONTENT}>
+                        <div className={classNames(CARD_HEADER_CONTENT, { [`${CARD_HEADER_CONTENT}--expandable`]: expandable })}>
+                            {expandable && (
+                                <div className={CARD_TOGGLE_CLASS}>
+                                    {showContent ? <ChevronUp role="presentation" /> : <ChevronDown role="presentation" />}
+                                </div>
+                            )}
                             {renderHeader ? renderHeader : <span className={CARD_TITLE}>{title}</span>}
                             {subTitle && <div className={CARD_SUBTITLE}>{subTitle}</div>}
                         </div>
```
glad-adyen, line 92: For improved accessibility, this `div` element should become the interaction trigger instead of the `section` element. Hence, all the extra props defined by the `cardContainerAttributes` object should be applied to this `div` element instead.

src/components/internal/Card/Card.scss
```diff
@@ -41,13 +45,25 @@
...
     &__header {
         border-bottom: style.token(color-outline-secondary) 1px solid;
+        cursor: pointer;
         display: block;
         padding: style.token(spacer-040) style.token(spacer-070) style.token(spacer-020);
         width: 100%;
+
+        &--expandable {
+            border-bottom: unset;
+        }
```
glad-adyen, line 55: Instead of unsetting `border-bottom` property, which could have unintended consequences, I'd recommend inversion with the `:not` pseudo-class and only specifying the `border-bottom` for the inverted rule.

```suggestion
        cursor: pointer;
        display: block;
        padding: style.token(spacer-040) style.token(spacer-070) style.token(spacer-020);
        width: 100%;

        &:not(&--expandable) {
            border-bottom: style.token(color-outline-secondary) 1px solid;
        }
```

glad-adyen (changes requested):

src/components/external/DisputeManagement/context/dispute/context.tsx
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }, [clearFiles, flowState]);
+
+    const clearStates = useCallback(() => {
+        setSelectedDefenseReason(null);
+        setApplicableDocuments(null);
+        setDispute(undefined);
+        setDefendDisputePayload(null);
+        setDefendResponse(null);
+    }, [setApplicableDocuments, setDispute, setSelectedDefenseReason]);
```
glad-adyen, line 74: Since all of these dependencies (except `setDispute`) are state setters, you can skip them here.
```suggestion
    }, [setDispute]);
```

src/components/external/DisputeManagement/context/dispute/context.tsx
```diff
@@ -0,0 +1,127 @@
...
+                    if (value instanceof File) {
+                        formData.set(field, value, value.name);
+                    } else formData.set(field, value);
+                }
+
+                formData.set(name, file, file.name);
+                return formData;
+            });
+        },
+        [setDefendDisputePayload]
```
glad-adyen, line 90: Since this dependency is a state setter, you can skip it here.
```suggestion
        []
```

glad-adyen (dismissed):

glad-adyen (approved):

PR #693, opened 2025-06-04
Fix defend reason dropdown

ElifDogaliBal:
## Summary
- Disable dropdown when there is only one defend reason
- Hide chevron when there is only one defend reason
- Fix the position of the defend reason dropdown
- Change the gap and make it 

**Fixed issue**:  [CXP-3727](https://youtrack.is.adyen.com/issue/CXP-3727/Fix-UI-of-new-Defense-flow)

### DIFF
--- a/src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
+++ b/src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
@@ -144,11 +144,12 @@
 }
 
 .adyen-pe-defend-dispute-document-upload-box {
-    @include adyen-pe-dropdown-overrides;
-
     border: 1px solid style.token(color-outline-primary);
     border-radius: style.token(border-radius-l);
     padding: style.token(spacer-080);
+
+    // stylelint-disable order/order
+    @include adyen-pe-dropdown-overrides;
 
     &__delete-button-container {
         align-items: center;
--- a/src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
+++ b/src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
@@ -3,7 +3,7 @@
 @mixin adyen-pe-dropdown-overrides {
     &__dropdown-list {
         max-height: 270px;
-        overflow: hidden scroll;
+        overflow: hidden auto;
 
         &--mobile {
             left: 0;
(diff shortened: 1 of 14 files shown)


src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
```diff
@@ -107,6 +107,19 @@
...
+        overflow: scroll;
+
+        &--mobile {
+            left: 0;
+            position: fixed;
+            top: 0;
+            width: 80%;
+            z-index: 30;
+        }
+    }
```
glad-adyen, line 122: Instead of duplicating this, what do you think about making this a mixin?

```scss
@mixin adyen-pe-dropdown-overrides {
    &__dropdown-list {
        max-height: 270px;
        overflow: scroll;

        &--mobile {
            left: 0;
            position: fixed;
            top: 0;
            width: 80%;
            z-index: 30;
        }
    }
}
```

```suggestion
    @include adyen-pe-dropdown-overrides;
```

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
```diff
@@ -133,6 +146,19 @@
...
```
glad-adyen, line 160: Ditto: previous [comment](https://github.com/Adyen/adyen-platform-experience-web/pull/693/files#r2144543342)

src/components/external/DisputeManagement/components/DefendDisputeFlow/DefendDisputeFlow.scss
```diff
@@ -107,6 +107,19 @@
         color: style.token(color-label-secondary);
         font-size: style.token(text-caption-font-size);
     }
+
+    &__dropdown-list {
+        max-height: 270px;
+        overflow: scroll;
```
glad-adyen, line 113: Since we are constraining just the height here, I'd say we explicitly use `overflow-y` instead of `overflow`. Or as an alternative, set the two axes overflow values using the shorthand syntax.

```suggestion
        overflow: hidden scroll;
```

src/components/internal/Popover/Popover.scss
```diff
@@ -12,7 +12,7 @@
         box-shadow: 0 6px 12px rgba(0, 17, 44, 0.08), 0 2px 4px rgba(0, 17, 44, 0.04), 0 0 0 1px rgba(0, 17, 44, 0.06);
         min-width: 200px;
         overflow: hidden;
-        z-index: 10;
+        z-index: 100;
```
glad-adyen, line 15: Is this ramp up really necessary? Are there challenges with the current stacking index?

isuert (dismissed):
🔧 ✨

glad-adyen (approved):

PR #739, opened 2025-08-13
Accessibility improvements for copy text component

glad-adyen:
## Summary
This PR introduces some changes to the `<CopyText />` for improved accessibility.

- Added live region for "Copied" status announcements
- Repurposed and renamed `buttonLabel` prop to `copyButtonAriaLabelKey`, to provide accessible name for copy button
- Added new `visibleText` prop to take on the previous behavior of the `buttonLabel` prop
- Refactored class names and render logic for better organization

**Addresses:**
- [piefp-15](https://adyen.hub.essentia11y.com/short-link/7f4NPBmmfukIz-Jv) — Ensure custom controls provide proper textual name, role, and state information
- [piefp-18](https://adyen.hub.essentia11y.com/short-link/Zc_40SUFjACcCAWE) — Ensure that status messages can be determined programmatically without receiving focus
- [piefp-20](https://adyen.hub.essentia11y.com/short-link/Ma3rn0sDBfdaTVux) — Ensure custom controls provide proper textual name, role, and state information
- [piefp-24](https://adyen.hub.essentia11y.com/short-link/MiBAbIHnCJGhkoj_) — Ensure that status messages can be determined programmatically without receiving focus
- [piefp-25](https://adyen.hub.essentia11y.com/short-link/IEXLXE8DLin60lex) — Ensure custom controls provide proper textual name, role, and state information
- [piefp-27](https://adyen.hub.essentia11y.com/short-link/TIHsK9vtPebOXLUs) — Ensure that status messages can be determined programmatically without receiving focus
- [piefp-42](https://adyen.hub.essentia11y.com/short-link/Yu4xAeP_idjaYZty) — Ensure custom controls provide proper textual name, role, and state information
- [piefp-49](https://adyen.hub.essentia11y.com/short-link/MCh0fsZs0rTr5X37) — Ensure that status messages can be determined programmatically without receiving focus
- [piefp-55](https://adyen.hub.essentia11y.com/short-link/VY4m1UkTgk-MnX4p) — Ensure custom controls provide proper textual name, role, and state information
- [piefp-59](https://adyen.hub.essentia11y.com/short-link/-ZKHerGNLSE_x3iJ) — Ensure that status messages can be determined programmatically without receiving focus

**Related to: [CXP-3361](https://youtrack.is.adyen.com/issue/CXP-3361)**


src/components/internal/CopyText/CopyText.tsx
```diff
@@ -16,74 +16,92 @@ type CopyTextProps = {
...
                         })}
                     >
                         <Icon name={'copy'} data-testid={'copy-icon'} />
                     </div>
                 </Button>
             </Tooltip>
+
+            <div className="adyen-pe-visually-hidden" aria-atomic="true" aria-live={isCopied ? 'polite' : 'off'}>
+                {isCopied && copiedStatusText}
+            </div>
```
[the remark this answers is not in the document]
glad-adyen: Got it

src/components/external/CapitalOverview/components/GrantItem/GrantItem.tsx
```diff
@@ -93,7 +93,8 @@ export const GrantItem: FunctionalComponent<GrantItemProps> = ({ grant, showDeta
                     <div className={GRANT_ITEM_CLASS_NAMES.grantID}>
                         <CopyText
                             textToCopy={grant.id}
-                            buttonLabel={i18n.get('capital.grantID')}
+                            visibleText={i18n.get('capital.grantID')}
+                            buttonLabelKey="capital.grantItem.copyGrantID"
```
ElifDogaliBal, line 97: What do you think about having the same strategy for translations of visibleText and buttonLabelKey to have consistency

buttonLabelKey={i18n.get('capital.grantItem.copyGrantID')}

glad-adyen: They serve different purposes and have different semantics, so maintaining the consistency here may not be the right call.

- `visibleText` — will be rendered on screen and is provided by the parent component, hence why it's translated from the parent component. It is rendered as received within the `<CopyText />` component, and as such, can be extended in the future to also render a `ComponentChild`. In summary, its localization is the responsibility of the parent component.

- `buttonLabelKey` — on the other hand, is only available to assistive technology (using `aria-label` attribute) as the accessible name of the copy icon button. As such, it has to be correctly localized, and the `<CopyText />` component guarantees same, by handling the `i18n.get(...)` call internally. In summary, its localization is the responsibility of the `<CopyText />` component itself.

Does this explanation make sense?

ElifDogaliBal: This semantically makes sense. On the other hand for other internal components like Button the parent component is sending the translations so, I think it might be easier to understand and maintain same structure. What do you think?

Another point what do you think about renaming buttonLabelKey as something like ariaLabelKey?

src/components/internal/CopyText/CopyText.tsx
```diff
@@ -1,89 +1,106 @@
...
+    label: BASE_CLASSNAME + '__label',
+    text: BASE_CLASSNAME + '__text',
+};
 
-    const [tooltipLabel, setTooltipLabel] = useState(i18n.get('copy'));
+const CopyText = ({ buttonLabelKey, isHovered, textToCopy, visibleText, showCopyTextTooltip = true, type = 'Link', ...restProps }: CopyTextProps) => {
+    const { i18n } = useCoreContext();
+    const [isCopied, setIsCopied] = useState(false);
+    const resetIsCopied = useCallback(() => setIsCopied(false), []);
+    const copyButtonLabel = useMemo(() => i18n.get(buttonLabelKey ?? 'copy'), [i18n, buttonLabelKey]);
```
ElifDogaliBal, line 38: Same as the previous command, what do you think about renaming copyButtonLabel to indicate it is only used for aria-label

```suggestion
    const ariaLabel = useMemo(() => i18n.get(buttonLabelKey ?? 'copy'), [i18n, buttonLabelKey]);
```

src/components/external/CapitalOverview/components/GrantItem/GrantItem.tsx
glad-adyen: > "...what do you think about renaming buttonLabelKey as something like ariaLabelKey?"

I'd prefer `buttonLabelKey` to `ariaLabelKey` because `ariaLabelKey` is too ambiguous (generic) and might be difficult to know what component it is associated with. Also, think of cases where there are multiple components that need accessibility name (label) from the author — in such cases, `ariaLabelKey` falls short from being a meaningful prop name.

ElifDogaliBal: For the first one, if that is the convention we are looking for, I suggest to create a ticket on youtrack to apply this to all internal components.

For the naming I think buttonLabelKey and copyButtonLabel doesn't suggest that it is only available to assistive technology so, at first glance it might be hard to guess its difference with visibleText. Since this is a prop name specifically for CopyText component I don't think it will create a confusion with other components. Instead of ariaLabelKey we can have something both makes it clear that it is only available to assistive technology and you think that it will not be ambiguous if you have any other suggestions.

ElifDogaliBal (dismissed):

PR #751, opened 2025-08-22
Send events to Mixpanel

ElifDogaliBal:
## Summary
- Add types for mixpanel requests
- Start sending events for Transactions Overview component


src/types/api/endpoints.ts
```diff
@@ -21,7 +23,9 @@ export type EndpointName = Extract<keyof EndpointsOperations, SetupResource['sch
 
 type CSVEndpoints = 'downloadReport';
 
-type JSONEndpoints = Exclude<EndpointName, CSVEndpoints | 'downloadDefenseDocument'>;
+type EncodedDataEndpoints = 'sendMixpanelEvent';
+
+type JSONEndpoints = Exclude<EndpointName, CSVEndpoints | 'downloadDefenseDocument' | EncodedDataEndpoints>;
```
glad-adyen, line 28: From the schema, `sendMixpanelEvent` returns a JSON response. Why is it excluded from the `JSONEndpoints` type here?

src/hooks/useAnalytics/usePushAnalyticEvent.ts
```diff
@@ -0,0 +1,33 @@
...
+        [sendAnalytics]
+    );
+
+    return (options: any) => {
+        const formattedOptions = JSON.stringify(options);
+        const encodedData = window.btoa(formattedOptions);
+        const data = new URLSearchParams();
+        data.append('data', encodedData);
+        track(data);
+    };
```
glad-adyen, line 32: Since this function is returned here, consider wrapping it in a `useCallback`, otherwise a new function will be created for each render.

src/hooks/useAnalytics/usePushAnalyticEvent.ts
```diff
@@ -0,0 +1,33 @@
...
+                }
+            ),
+        [sendAnalytics]
+    );
+
+    return (options: any) => {
+        const formattedOptions = JSON.stringify(options);
+        const encodedData = window.btoa(formattedOptions);
+        const data = new URLSearchParams();
+        data.append('data', encodedData);
```
glad-adyen, line 30: I'd recommend using `data.set()` method over `data.append()` here to ensure the parameter is completely overridden. I believe that is more consistent with the API expectation.

src/hooks/useAnalytics/usePushAnalyticEvent.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export const usePushAnalyticEvent = () => {
+    const { sendMixpanelEvent } = useConfigContext().endpoints;
+
+    const { mutate: sendAnalytics } = useMutation({
+        queryFn: sendMixpanelEvent,
+    });
+
+    const track = useCallback(
+        (options: any) =>
```
glad-adyen, line 13: Instead of leaving the type of the `options` parameter here as `any`, I'd recommend constraining its type to `URLSearchParams` (at least for now).

src/hooks/useAnalytics/useAnalytics.ts
```diff
@@ -0,0 +1,50 @@
+import { useCallback, useEffect } from 'preact/hooks';
+
```
glad-adyen, line 2: Not important (you can remove this empty line)

src/core/Http/utils.ts
```diff
@@ -47,6 +47,8 @@ export const getRequestBodyForContentType = (body: any, contentType: string) =>
             return JSON.stringify(body);
         case 'multipart/form-data':
             return body instanceof FormData ? body : new FormData();
+        case 'application/x-www-form-urlencoded':
+            return body;
```
glad-adyen, line 51: This seems redundant, considering the `default` case already ensures that the body is provided as a string (which is expected for `application/x-www-form-urlencoded`). Is that not the case?

glad-adyen (changes requested):

src/hooks/useBalanceAccountSelection.ts
```diff
@@ -49,9 +52,20 @@ const useBalanceAccountSelection = (balanceAccounts?: IBalanceAccountBase[], all
...
+                        category: 'Transaction component',
+                        value: balanceAccountId,
+                    });
+                }
+                isDefaultValueSelected.current = true;
+                setSelectedBalanceAccountIndex(index!);
+            }
         },
-        [allBalanceAccounts]
+        [allBalanceAccounts, userEvents, isDefaultValueSelected, selectedBalanceAccountIndex]
```
glad-adyen, line 68: You don't need a `ref` in the dependencies array — refs are not reactive.

src/core/Context/analytics/AnalyticsProvider.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export interface AnalyticsProviderProps {
+    componentName?: string;
+    children?: ComponentChildren;
+}
+
+/**
+ * The provider exists to scope analytics to one application instance.
+ * This would otherwise be problematic when a consumer might have several
+ * instantiations of a embedded Component for instance on the same page
```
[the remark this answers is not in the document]
ElifDogaliBal: This comment was remaining from kyc components project, this was related with a problem that they had before when they didn't implemented using context. This doesn't apply our project, removed the comment.

german664 (approved):

PR #764, opened 2025-09-15
Explicit headings

glad-adyen:
## Summary

**Fixed issue**:

### DIFF
--- a/src/components/internal/Header/Header.tsx
+++ b/src/components/internal/Header/Header.tsx
@@ -26,7 +26,7 @@ export interface HeaderProps {
 export const Header: FC<HeaderProps> = ({ baseClassName = BASE_CLASS, children, hasDivider, hideTitle, titleKey, subtitleKey, subtitleConfig }) => {
     const { i18n } = useCoreContext();
     return (
-        <header className={baseClassName}>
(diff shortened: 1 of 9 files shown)


src/components/internal/Header/Header.tsx
```diff
@@ -1,17 +1,19 @@
...
-import './Header.scss';
 import { Divider } from '../Divider/Divider';
+import './Header.scss';
 
 export const BASE_CLASS = 'adyen-pe-header';
 
-export interface HeaderProps {
+export interface HeaderProps extends HTMLProps<HTMLHeadingElement> {
     baseClassName?: string;
+    connected?: boolean;
```
isuert, line 16: What the meaning of this prop? Any alternative name that would convey the meaning better?

glad-adyen: With this description, do you have suggestions for a better name, or shall we resolve this?

isuert: Clear! WDYT of `relativeToRoot` or `relativeToRootComponent`? We can also complement this with some JSDoc

isuert (approved):

PR #774, opened 2025-10-02
Unit tests for useTimezoneAwareDateFormatting hook

glad-adyen:
## Summary
This PR introduces unit tests for the `useTimezoneAwareDateFormatting` hook.

### DIFF
--- a/src/hooks/useTimezoneAwareDateFormatting.test.ts
+++ b/src/hooks/useTimezoneAwareDateFormatting.test.ts
@@ -1,10 +1,173 @@
 /**
  * @vitest-environment jsdom
  */
-import { test } from 'vitest';
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
+import { renderHook } from '@testing-library/preact';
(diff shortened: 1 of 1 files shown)


src/components/internal/Calendar/hooks/useTimezone.test.ts
```diff
@@ -39,25 +39,26 @@ const validTimezones = [
...
         vi.useFakeTimers();
-        // Date.now() -> Dec 25, 2023, 4:10:45.123 PM UTC
-        vi.setSystemTime(systemTimestamp);
+        vi.setSystemTime(1703520645123); // Dec 25, 2023, 4:10:45.123 PM UTC
     });
 
     afterEach(() => {
         vi.useRealTimers();
         vi.restoreAllMocks();
+        clockUnsubscribes.length = 0;
```
german664, line 48: Should we reset the reference entirely instead of mutating the array?

glad-adyen: In this case, we are just clearing out the content of the array (which is sufficient).

german664 (approved):

PR #854, opened 2025-12-10
[DRAFT] Add risk module

german664:


.github/actions/setup-env/action.yml
```diff
@@ -46,7 +46,7 @@ runs:
     - name: Install dependencies
       shell: bash
       run: |
-        npm config set //registry.npmjs.org/:_authToken ${{ inputs.NPM_READ_TOKEN }}
+        npm config set //registry.npmjs.org/:_authToken ${{ inputs.NPM_PLATFORM_EXPERIENCE_TOKEN }}
```
glad-adyen: Why are we changing the token source here?

PR #1034, opened 2026-03-05
Embed onboarding components in capital

isuert:
## Summary
- Embeds onboarding components in capital flow
- Determines whether hosted or embedded onboarding will be used, based on the response from onboarding config endpoint
- Uses new analytic events for embedded onboarding flow
- Refactors the logic for rendering grant actions and unifies the copies used in the alerts


src/components/external/CapitalOverview/components/GrantActionsEmbedded/GrantActionsEmbedded.tsx
```diff
@@ -0,0 +1,162 @@
...
+    missingActions,
+}) => {
+    const { i18n, environment } = useCoreContext();
+    const userEvents = useAnalyticsContext();
+    const { getOnboardingConfiguration } = useConfigContext().endpoints;
+
+    const fetchToken = useCallback(async () => {
+        const data = await getOnboardingConfiguration?.(EMPTY_OBJECT);
+        return { token: data?.token || '' };
+    }, [getOnboardingConfiguration]);
```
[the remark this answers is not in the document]
isuert: The token that we pass from above would be outdated after a while, fetchToken ensures we get a fresh token every time it's needed.

mocks/mock-server/capital.ts
```diff
@@ -316,9 +298,9 @@ export const CapitalOverviewMockedResponses = capitalFactory({
         { endpoint: mockEndpoints.dynamicOfferConfig, response: EMPTY_OFFER },
         { endpoint: mockEndpoints.grants, response: { data: GRANTS } },
     ],
-    errorMissingActionsGeneric: [
+    errorActionsHosted: [
```
german664, line 301: This scenario doesn't include an `onboardingConfiguration` endpoint handler like others: 

```
{ endpoint: mockEndpoints.onboardingConfiguration, response: ONBOARDING_CONFIGURATION }
```

Without this mock, the request is unhandled by MSW, leading to non deterministic behavior and a possible flaky or failed test if default handler of `onboardingConfiguration` changes.

Should we add the handler to this one?

mocks/mock-server/utils/mocksHandlerFactory.ts
```diff
@@ -19,7 +19,7 @@ export const getHandlerCallback = <T extends JsonBodyType>({
         if (networkError) {
             return HttpResponse.error();
         }
-        await delay(delayTime ?? 0);
+        await delay(delayTime ?? 200);
```
german664, line 22: Why are we changing the default to 200?. It was `0` so we don't experience delay in mocks unless we explicitly send `delayTime`. Can you add the delay in the specific handlers you need it?

isuert: Since we use mocks for testing i think it's nice to add a small delay by default, so we can always see the loading states. WDYT?

mocks/mock-server/capital.ts
isuert: Good catch!

mocks/mock-server/utils/mocksHandlerFactory.ts
german664: Agree!

german664 (dismissed):

german664 (approved):

PR #1046, opened 2026-03-11
Make the shared Core framework-agnostic

german664:
## Summary

A first attempt to decouple `src/core/` from Preact by moving framework-specific implementations into `preact/` subdirectories and converting original files into re-export shims, so there are none consumer import changes.

Adapter split:

- CoreContext, CoreProvider, useCoreContext → src/core/Context/preact/
- AnalyticsContext, AnalyticsProvider, useAnalyticsContext → src/core/Context/analytics/preact/
- ConfigProvider, useConfigContext → src/core/ConfigContext/preact/

Shared setup functions (pure TS, reusable by any future framework adapter):

- setupAnalytics.ts: user events factory + subscribe/unsubscribe lifecycle
- setupConfig.ts: session subscription, permission checking, context value derivation
- setupCore.ts: i18n readiness gate, context value assembly

Type cleanup:

- Replaced Preact RefObject with generic ComponentRef interface
- Made interpolateElement framework-agnostic (generic type parameter)
- Removed all preact imports from non-adapter files in src/core/ (except deferred ErrorPanel.tsx)

## Tested scenarios

All unit and integration tests pass.

**Fixed issue**: [IEX-2294](https://youtrack.is.adyen.com/issue/IEX-2294)


src/core/Context/preact/CoreProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { CoreProviderProps } from '../types';
+import useBooleanState from '../../../hooks/useBooleanState';
+import { waitForI18n, createCoreContextValue } from '../setupCore';
+import Localization from '../../Localization';
+
+/**
+ * CoreProvider Component
+ * Wraps a component delaying the render until after the i18n module is fully loaded
+ */
+const CoreProvider = (props: CoreProviderProps) => {
```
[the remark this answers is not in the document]
german664: ✅

src/core/Context/preact/CoreProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+    const { children, i18n = new Localization().i18n } = props;
+    const [ready, setReady] = useBooleanState(false);
+
+    useEffect(() => {
+        waitForI18n(i18n)
+            .then(() => setReady(true))
+            .catch();
```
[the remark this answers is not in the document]
german664: ✅

src/core/Context/preact/CoreProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+    }, []);
+
+    const coreContextValues = useMemo(
+        () => createCoreContextValue({ ...props, i18n }),
```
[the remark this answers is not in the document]
german664: ✅

src/core/Context/preact/CoreProvider.tsx
```diff
@@ -0,0 +1,44 @@
...
+        [
+            props.commonProps,
+            props.componentRef,
+            props.externalErrorHandler,
+            i18n,
```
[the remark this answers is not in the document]
german664: ✅

src/core/Context/CoreProvider.tsx
```diff
@@ -66,4 +66,4 @@ const CoreProvider = ({
     return <CoreContext.Provider value={coreContextValues}>{toChildArray(children)}</CoreContext.Provider>;
 };
```
samuelsa-adyen: This has been moved, so it can be removed.

src/core/Context/analytics/setupAnalytics.ts
```diff
@@ -0,0 +1,32 @@
...
+    userEvents: Partial<UserEvents>;
+    subscribe(pushEvent: (data: EventQueueItem) => void): () => void;
+}
+
+export function setupAnalytics({ analyticsEnabled, componentName }: AnalyticsSetupOptions): AnalyticsSetupResult {
+    const userEvents = createUserEvents(analyticsEnabled, componentName);
+
+    userEvents.updateBaseTrackingPayload?.({
+        sdkVersion: process.env.VITE_VERSION,
+        userAgent: navigator.userAgent,
```
samuelsa-adyen: Not sure if the components will be rendered in SSR, but the navigator might be undefined.

```suggestion
        userAgent: typeof navigator !== 'undefined' ? navigator.userAgent : '',
```

src/core/Context/preact/CoreProvider.tsx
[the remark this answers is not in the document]
samuelsa-adyen: Maybe we could consider adding some analytics/error messaging in this catch to know if this ever happens.

src/core/Context/types.ts
```diff
@@ -14,7 +17,7 @@ export interface CoreProviderProps {
     loadingContext?: string;
     updateCore?: () => void;
     externalErrorHandler?: onErrorHandler | null;
-    componentRef: RefObject<HTMLDivElement>;
+    componentRef: ComponentRef<HTMLDivElement>;
     getImageAsset?: (props: AssetOptions) => string;
     getDatasetAsset?: (props: AssetOptions) => string;
     getCdnConfig?: <Fallback>(props: { name: string; extension?: string; subFolder?: string; fallback?: Fallback }) => Promise<Fallback>;
```
glad-adyen, line 21: This interface still exhibits some P(r)eact specific characteristics — e.g. `children` and `componentRef` (having a `current` property).

src/core/Context/preact/useCoreContext.ts
```diff
@@ -0,0 +1,6 @@
+import { useContext } from 'preact/hooks';
+import { CoreContext } from '../CoreContext';
```
glad-adyen: I'd recommend importing the `CoreContext` directly from the `preact` directory here, instead of the parent directory. The parent should only be responsible for remapping exports to the Preact variants.

```suggestion
import { CoreContext } from './CoreContext';
```

src/core/Context/CoreProvider.tsx
glad-adyen: Indeed — this module should only re-export Preact's variant of the `CoreProvider`.

src/core/Context/preact/CoreProvider.tsx
```diff
@@ -0,0 +1,39 @@
+import { toChildArray } from 'preact';
+import { useEffect, useMemo } from 'preact/hooks';
+import { CoreContext } from '../CoreContext';
```
glad-adyen: I'd recommend importing the `CoreContext` directly from the `preact` directory here, instead of the parent directory. The parent should only be responsible for remapping exports to the Preact variants.

```suggestion
import { CoreContext } from './CoreContext';
```

src/core/Context/preact/CoreProvider.tsx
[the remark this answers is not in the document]
glad-adyen: Actually, we don't need the `catch()`, because `waitForI18n()` will never throw. Any potential error is handled upstream and logged using `console.error()`.

```suggestion
    useEffect(() => {
        waitForI18n(props.i18n)
            .then(() => setReady(true));
```

src/core/Context/analytics/preact/useAnalyticsContext.tsx
```diff
@@ -0,0 +1,15 @@
+import { useContext } from 'preact/hooks';
+
+import { AnalyticsContext } from '../AnalyticsContext';
```
glad-adyen:
```suggestion
import { AnalyticsContext } from './AnalyticsContext';
```

src/core/Context/analytics/preact/AnalyticsProvider.tsx
```diff
@@ -0,0 +1,35 @@
+import type { ComponentChildren } from 'preact';
+import { PropsWithChildren } from 'preact/compat';
+import { useCallback, useEffect, useMemo } from 'preact/hooks';
+import { AnalyticsContext } from '../AnalyticsContext';
```
glad-adyen:
```suggestion
import { AnalyticsContext } from './AnalyticsContext';
```

src/core/Context/analytics/setupAnalytics.ts
glad-adyen: Indeed.

We should consider something more robust, since our components are not designed to be rendered fully server-side (of course, importing them dynamically will be fine). Here are some thoughts:

- State in the public docs never to render the components server-side (no SSR). And if SSR must be used, the components should be dynamically loaded.

- Add an environment to the `core` instance initialization function that warns when you are initializing from server-side (in development mode, only when using NPM package).

```ts
if (typeof window === 'undefined' || typeof window.document === 'undefined') {
    // server-side
    console.warn(...);
}
```

src/core/ConfigContext/preact/context.tsx
```diff
@@ -0,0 +1,48 @@
...
+                        centered
+                        title={'common.errors.somethingWentWrong'}
+                        message={[componentAvailabilityErrors(type), 'common.errors.contactSupport']}
+                    />
+                ))}
+        </ConfigContext.Provider>
+    );
+};
+
+export const useConfigContext = () => useContext(ConfigContext);
```
glad-adyen, line 47: Would be nice to introduce a little bit of defensive programming here (consistent with other context hooks).


```suggestion
export const useConfigContext = () => {
    const context = useContext(ConfigContext);

    if (!context) {
        throw new Error('Cannot use ConfigContext without <ConfigProvider>');
    }

    return context;
};

```

src/core/Context/CoreProvider.tsx
german664: Nice catch, removed ✅

src/core/ConfigContext/preact/context.tsx
samuelsa-adyen: Would there ever not be a context tho? We're currently setting the context with default values, so from the definition we would have an object in it.

glad-adyen: Indeed — More for consistency with other context hooks than defense against broken flow (based on current setup)

src/core/Context/analytics/setupAnalytics.ts
german664: Good catch. I tried to follow Glad's suggestion to address this more broadly: analytics now use safe accessors for SSR and core initialization emits a warning in DEV when invoked without window/document. This should avoid server side crashes and make more explicit that SSR isn’t supported and components should be dynamically loaded in SSR setups.

src/core/ConfigContext/preact/context.tsx
samuelsa-adyen: Could we standardize then just having `export const useConfigContext = () => useContext(ConfigContext);` instead? Seems unnecessary to have a condition that will never get executed.

src/core/Context/types.ts
german664: Refactored `ComponentRef` from `{ current: T | null }` to a getter function `() => T | null` (so we can wrap Vue's own ref mechanism into a simple getter). 

Kept `children?: any` because it's typed as `any` with no Preact imports, so it's already framework-agnostic. During the Vue migration we could map slots to this prop or remove it entirely, depending on the needs in the future.

src/core/ConfigContext/preact/context.tsx
german664: I agree with Sam. Since all contexts are initialized with default values, the null check is unreachable, so I'm thinking of keeping the export as is. 

If consistency is a concern, I think we can clean up the guards in the other context hooks in a follow up PR.

src/components/external/UIElement/UIElement.tsx
```diff
@@ -87,7 +87,7 @@ export class UIElement<P> extends BaseElement<P & UIElementProps> implements IUI
         return (
             <ConfigProvider type={this.type} session={core.session} key={performance.now()}>
                 <CoreProvider
-                    componentRef={this.compRef}
+                    componentRef={() => this.compRef.current}
```
glad-adyen: Seeing how the downstream usage of this callback, I recommend making it more stable.

First define it as a `const` outside the Preact render tree (but still within the `render` function). This will ensure referential stability of the callback.
```ts
const componentRefGetter = () => this.compRef.current;
```

Then you can use the callback within the render tree like so:
```tsx
<CoreProvider
    componentRef={componentRefGetter}
```

german664: Great suggestion! 

✅

samuelsa-adyen (approved):

glad-adyen (approved):

PR #1048, opened 2026-03-13
refactor(testing): add stable Playwright test hooks

german664:
## Summary
Adds framework-agnostic test hooks to shared UI primitives and updates shared Playwright helpers/models to consume them.

## Why
The current integration tests rely on framework-coupled selectors in several shared paths. This PR introduces stable `data-testid` and accessible hook points so tests can target component behavior rather than implementation details.

## Included
- stable test hooks in shared/internal components
- targeted test hooks in shared external components used by tests
- helper/model updates in `tests/utils` and `tests/models`

## Validation
- `pnpm run types:check`
- targeted Playwright coverage for updated shared paths


tests/models/internal-components/dataGrid.ts
```diff
@@ -4,13 +4,15 @@ class DataGridPage {
...
-    constructor(page: Page, rootElementSelector: string) {
-        this.rootElement = page.locator(rootElementSelector).getByRole('table');
+    constructor(page: Page, rootElementSelector?: string) {
+        const rootElement = rootElementSelector ? page.locator(rootElementSelector) : page.getByTestId('component-root');
+
+        this.rootElement = rootElement.getByRole('table');
         this.gridBody = this.rootElement.getByRole('rowgroup').nth(1);
     }
 
     getCell(label: string, row = 0) {
```
glad-adyen: I would recommend changing the name of the parameter here from `label` to `testId` to better reflect how it is being used.

```suggestion
    getCell(testId: string, row = 0) {
```

german664: ✅

tests/models/internal-components/dataGrid.ts
```diff
@@ -4,13 +4,15 @@ class DataGridPage {
...
+        const rootElement = rootElementSelector ? page.locator(rootElementSelector) : page.getByTestId('component-root');
+
+        this.rootElement = rootElement.getByRole('table');
         this.gridBody = this.rootElement.getByRole('rowgroup').nth(1);
     }
 
-    getCell(label: string, row = 0) {
-        return this.getRow(row).locator(`div[aria-labelledby=${label}]`);
+    getCell(testId: string, row = 0) {
+        return this.getRow(row).getByTestId(label);
```
glad-adyen: Now we missed switching to the `testId` parameter in the body of the function.


```suggestion
        return this.getRow(row).getByTestId(testId);
```

src/components/internal/FormWrappers/FormField.tsx
```diff
@@ -10,11 +10,12 @@ const FormField = ({
...
     children,
     className,
-}: PropsWithChildren<{ label: string; optional: boolean; supportText?: string; className?: string }>) => {
+    testId,
+}: PropsWithChildren<{ label: string; optional: boolean; supportText?: string; className?: string; testId?: string }>) => {
     const { i18n } = useCoreContext();
 
     return (
-        <div className={cx('adyen-pe-payment-link-creation-form__field-container', className)}>
+        <div className={cx('adyen-pe-payment-link-creation-form__field-container', className)} role="group" aria-label={label} data-testid={testId}>
```
glad-adyen: Why is this modified to have a `"group"` role here? The `"group"` role is particularly useful for defining a logical group of related interactive elements. Since it might be a bit tricky to establish the content of `children`, it is better to remove it.


```suggestion
        <div className={cx('adyen-pe-payment-link-creation-form__field-container', className)} data-testid={testId}>
```

src/components/internal/FormWrappers/FormSelect.tsx
```diff
@@ -40,6 +40,7 @@ export function FormSelect<TFieldValues>({
     validate,
 }: FormSelectProps<TFieldValues>) {
     const { control, fieldsConfig, getValues, setValue } = useWizardFormContext<TFieldValues>();
+    const fieldNameValue = String(fieldName);
```
glad-adyen: Why is this explicit `String()` casting needed here?

Using the `fieldName` value directly in the template literal (like so: `form-field-${fieldName}` or `field-error-${fieldName}`) implicitly has the same effect as casting to string.

Also, would be nice to double check  and ensure that the `fieldName` prop is always a non-empty string. Otherwise, it would result in malformed and less useful test ids.

src/components/internal/FormWrappers/FormTextInput.tsx
```diff
@@ -49,6 +49,7 @@ export function FormTextInput<TFieldValues>({
 }: FormTextInputProps<TFieldValues>) {
     const { control, fieldsConfig } = useWizardFormContext<TFieldValues>();
     const { i18n } = useCoreContext();
+    const fieldNameValue = String(fieldName);
```
glad-adyen: Same here. See other [comment](https://github.com/Adyen/adyen-platform-experience-web/pull/1048/changes#r2988063755)

src/components/internal/StructuredList/StructuredList.tsx
```diff
@@ -58,7 +58,7 @@ export default function StructuredList({
                             <Typography variant={typographyVariant}>{item.label}</Typography>
                         )}
                     </dt>
-                    <dd className={cx(SL_CONTENT_CLASS, VALUE_COL_CLASS)}>
+                    <dd className={cx(SL_CONTENT_CLASS, VALUE_COL_CLASS)} data-testid={`${item.key}-value`}>
```
glad-adyen: Just thinking, `item.key` may not be the ideal choice for constructing test ids, since it is usually a translation key.

german664: ✅ Switched to `item.id` with a guard, so the test ID is only set when an identifier is available!

src/components/internal/FormWrappers/FormTextInput.tsx
german664: ✅

src/components/internal/FormWrappers/FormSelect.tsx
german664: ✅

src/components/internal/FormWrappers/FormField.tsx
german664: ✅

.factory/settings.json
glad-adyen, line 1: Did you intend to commit this file?

german664: Removed! ✅

glad-adyen (approved):

samuelsa-adyen (approved):

PR #1050, opened 2026-03-13
chore(testing): enforce Playwright selector guard

german664:
## Summary
Adds an ESLint-based guardrail that blocks framework-coupled Playwright selectors from being introduced in specs and page models.

## Why
After migrating the tests, we want enforcement in the normal lint flow so the old selector patterns do not regress back into the suite.

## Included
- custom ESLint rule for disallowed Playwright selector patterns
- dedicated ESLint config for Playwright specs/models
- `package.json` lint wiring for the new check

## Validation
- `pnpm run lint:playwright-selectors`
- synthetic rule verification against a forbidden selector
- `pnpm run types:check`


package.json
```diff
@@ -54,7 +54,8 @@
     "build:publish": "pnpm run build && pnpm pack",
     "changeset": "changeset",
     "format": "prettier --write 'src/**/*.{ts,js,css,scss,html}'",
-    "lint": "ESLINT_USE_FLAT_CONFIG=false eslint 'src/**/*.{js,ts,tsx}' --quiet && stylelint 'src/**/*.scss' --fix",
+    "lint": "ESLINT_USE_FLAT_CONFIG=false eslint 'src/**/*.{js,ts,tsx}' --quiet && stylelint 'src/**/*.scss' --fix && pnpm run lint:playwright-selectors",
+    "lint:playwright-selectors": "eslint --no-config-lookup --config scripts/eslint-playwright-selectors.cjs 'tests/**/*.{spec,test}.{ts,tsx}' 'tests/models/**/*.{ts,tsx}'",
```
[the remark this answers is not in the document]
german664: Not needed. We don't write js, just ts.

scripts/eslint-rules/no-framework-coupled-playwright-selectors.cjs
```diff
@@ -0,0 +1,103 @@
...
+        regex: /input\[name=/,
+    },
+    {
+        label: 'name attribute selectors for divs',
+        regex: /div\[name=/,
+    },
+    {
+        label: 'title attribute selectors for buttons',
+        regex: /button\[title=/,
+    },
```
glad-adyen, line 32: Instead of using specific attribute selectors here, combined with a small set of elements, maybe it might be worth disallowing attribute selectors in general. This will force tests to use more high-priority locator strategies such as role-based or test-id locators.

scripts/eslint-rules/no-framework-coupled-playwright-selectors.cjs
```diff
@@ -0,0 +1,103 @@
...
+        label: 'internal CSS class selectors',
+        regex: /\.adyen-pe-/,
+    },
+    {
+        label: 'XPath selectors',
+        regex: /xpath=/,
+    },
+    {
+        label: 'definition list structure selectors',
+        regex: /^(?:dd|dl)$/,
```
glad-adyen: This is also too specific. Would you consider being more generic with this restriction, like to prevent direct element selectors in general?

german664: ✅ Changed it with a more generic bare type selector regex

scripts/eslint-rules/no-framework-coupled-playwright-selectors.cjs
german664: I agree, but I think a full attribute selector ban may be a bit too much _right now_ because we still rely on some `data-*` hooks and cases like `input[type="file"]`. I’d prefer banning the more clear fragile attributes first, and then tightening more (perhaps even all attribute selectors) once the remaining tests have better role/test-id coverage.

glad-adyen (approved):

PR #1052, opened 2026-03-17
Introduce font family and locale as global story args

isuert:
## Summary
Introducing a way to test our stories with a distinct font family (comic sans in this case). This helps in visually verifying that font family is consistent across our components and onboarding ones.


stories/utils/helpers.ts
```diff
@@ -1,5 +1,15 @@
 import { action } from 'storybook/actions';
 
+export const commonArgTypes = {
```
german664: I think instead of this, we can define in `.storybook/preview.tsx` a "globalTypes", something like:

```
globalTypes: {
        fontFamily: {
        control: { type: 'select' },
        options: ['Default', 'Comic Sans MS'],
        mapping: {
            Default: undefined,
        },
    },
```


This will add a new control in the toolbar. Is a storybook native way of doing it and we won't have the need to import this in every story. Just consume it in the Container like `context.globals.fontFamily`

stories/utils/Container.tsx
```diff
@@ -28,19 +28,24 @@ export const Container = <T extends new (args: any) => any>({ component, compone
...
+    }, [component, componentConfiguration, context, locale]);
 
     return (
         <>
-            <div ref={container} id="component-root" className="component-wrapper" />
+            <div
+                ref={container}
+                id="component-root"
+                className="component-wrapper"
+                style={{ fontFamily: context.args.fontFamily as string | undefined }}
```
german664: Here you could do something like `{{ fontFamily: context.globals.fontFamily as string | undefined }}` based on https://github.com/Adyen/adyen-platform-experience-web/pull/1052/changes#r2947786313

stories/utils/Container.tsx
```diff
@@ -28,19 +28,24 @@ export const Container = <T extends new (args: any) => any>({ component, compone
...
                 },
-                ...context.args.coreOptions,
+                ...(context.args.coreOptions ?? {}),
             });
 
             Component = new component({ ...componentConfiguration, core });
             Component.mount(container.current ?? '');
         })();
 
         return () => Component.unmount();
```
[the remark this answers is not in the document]
german664: This has small impact but I still think we can apply this suggestion @isuert

german664 (approved):

samuelsa-adyen (approved):

.storybook/preview.tsx
```diff
@@ -12,6 +12,28 @@ const preview: Preview = {
...
+                title: 'Locale',
+                items: ['da-DK', 'de-DE', 'en-US', 'es-ES', 'fi-FI', 'fr-FR', 'it-IT', 'nl-NL', 'no-NO', 'pt-BR', 'sv-SE'],
+                dynamicTitle: true,
+            },
+        },
+        fontFamily: {
+            description: 'Global font family for components',
+            toolbar: {
+                title: 'Font family',
+                items: [{ title: 'Default', value: undefined }, { value: 'Comic Sans MS' }],
```
samuelsa-adyen, line 28: Superb choice 👍🏼

stories/utils/Container.tsx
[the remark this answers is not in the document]
isuert: it's applied

PR #1059, opened 2026-03-20
Change onboarding version

isuert:


package.json
```diff
@@ -75,7 +75,7 @@
     "version": "./scripts/bump-version.sh"
   },
   "dependencies": {
-    "@adyen/kyc-components": "4.0.0-compact.4",
+    "@adyen/kyc-components": ">=4.0.0-compact.3 <4.0.0",
```
samuelsa-adyen, line 78: We can simplify this to 
```suggestion
    "@adyen/kyc-components": "^4.0.0-compact.4",
```

isuert: That's not going to work - we don't want 4.0.1+.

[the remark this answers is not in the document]
isuert: We use version ranges for our dependecies and pnpm-lock.

samuelsa-adyen (approved):

PR #1065, opened 2026-03-27
fix: eslint react-hooks/exhaustive-deps remaining

glad-adyen:
## Summary
This PR introduces changes to fix the remaining ESLint `react-hooks/exhaustive-deps` rule violations across the entire project, with a few exceptions (which otherwise, break the existing behavior in unintended ways).

**Fixed issue: [IEX-2399: Fix react-hooks/exhaustive-deps violations](https://youtrack.is.adyen.com/issue/IEX-2399)**

### DIFF
--- a/src/components/internal/Calendar/hooks/useCalendar.ts
+++ b/src/components/internal/Calendar/hooks/useCalendar.ts
@@ -74,6 +74,7 @@ const useCalendar = (
         };
 
         return { grid, kill };
+        // eslint-disable-next-line react-hooks/exhaustive-deps
     }, []);
 
     const cursorRootProps = useMemo(() => {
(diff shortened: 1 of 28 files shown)


src/components/internal/Calendar/hooks/useCalendar.ts
```diff
@@ -74,6 +74,7 @@ const useCalendar = (
         };
 
         return { grid, kill };
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
german664, line 77: Why are we disabling the eslint rule instead of adding the missing deps?

glad-adyen: Because, the outcome causes too many unnecessary re-renders, which breaks the existing behavior in unintended ways. For these kinds of cases, the rule is disabled.

german664 (dismissed):

samuelsa-adyen (dismissed):

german664 (approved):

samuelsa-adyen (approved):

PR #1067, opened 2026-03-27
fix: eslint react-hooks/purity

glad-adyen:
## Summary
This PR introduces changes to fix the existing ESLint `react-hooks/purity` rule violations across the entire project.

**Fixed issue: [IEX-2402: Fix react-hooks/purity violations](https://youtrack.is.adyen.com/issue/IEX-2402)**

### DIFF
--- a/src/components/internal/DatePicker/components/TimeRangeSelector/useTimeRangeSelection.ts
+++ b/src/components/internal/DatePicker/components/TimeRangeSelector/useTimeRangeSelection.ts
@@ -54,7 +54,7 @@ export const getTimeRangeSelectionDefaultPresetOptions = ({
 };
 
 export const useTimeRangeSelection = ({
-    now = Date.now(),
+    now,
     options: presetOptions,
     selectedOption: selectedPresetOption,
     timezone,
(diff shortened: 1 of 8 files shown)


german664 (approved):

german664 (approved):

src/components/internal/DatePicker/components/TimeRangeSelector/useTimeRangeSelection.ts
```diff
@@ -116,7 +116,7 @@ export const useTimeRangeSelection = ({
             const options = Object.values(presetOptions);
 
             options.forEach(ranges => {
-                ranges.now = now;
+                ranges.now = now ?? null;
```
samuelsa-adyen, line 120: This will change the behavior, no? we're switching from `Date.now()` as the default to `null`

glad-adyen: The change isn't exactly 1:1. But the resulting behavior is the same, because the range objects default to `Date.now()` when `now` is set to a nullish value (like `null` in this case).

samuelsa-adyen (approved):

PR #1070, opened 2026-03-27
Chore/nx publish contract

german664:
## Summary

This PR sets the initial publish-contract safeguards for the package extraction effort. It introduces the initial Nx boundary taxonomy and a publish-diff baseline/check/test flow to catch unintended public export, type declaration, and package entrypoint changes during the migration.


scripts/check-publish-contract/check.ts
samuelsa-adyen, line 1: What do we thing about writing these with TypeScript running them with tsx instead to get type safety?

scripts/publish-diff/check.mjs
```diff
@@ -0,0 +1,57 @@
...
+    const baseline = JSON.parse(readFileSync(BASELINE_PATH, 'utf-8'));
+    const diffs = diff(baseline, snapshot);
+
+    if (diffs.length === 0) {
+        console.log('Publish contract: no changes detected');
+        process.exit(0);
+    }
+
+    console.error('Publish contract CHANGED:');
+    for (const entry of diffs) console.error(entry);
```
samuelsa-adyen: I think there was an eslint rule that required us to add brackets for conditionals


```suggestion
    for (const entry of diffs) {
      console.error(entry);
    }
```

scripts/check-publish-contract/check.mjs
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    console.error('Publish contract CHANGED:');
+    for (const entry of diffs) console.error(entry);
+    console.error('\nIf intentional, update the baseline:');
+    console.error('  pnpm run publish-diff -- --update');
+    process.exit(1);
+}
+
+const isExecutedDirectly = !!process.argv[1] && resolve(process.argv[1]) === fileURLToPath(import.meta.url);
```
samuelsa-adyen: What do we use this for?

package.json
```diff
@@ -54,6 +54,8 @@
     "build:publish": "pnpm run build && pnpm pack",
     "changeset": "changeset",
     "format": "prettier --write 'src/**/*.{ts,js,css,scss,html}'",
+    "publish-diff": "pnpm run build && node scripts/publish-diff/check.mjs",
```
samuelsa-adyen: `publish-diff` sounds like we are going to publish something, maybe we could change this to something around the lines of `check-publish-contract` or something like it

german664: Agree! Changed ✅

scripts/check-publish-contract/check.mjs
german664: Not really needed, so removed!

scripts/check-publish-contract/check.ts
german664: ✅

samuelsa-adyen (approved):

PR #1082, opened 2026-03-31
fix: eslint unused vars and generic violations

glad-adyen:
## Summary
This PR introduces changes to fix the existing violations of the following ESLint rules across the entire project, with a few exceptions (which otherwise, might break the existing behavior in unintended ways).

- `@typescript-eslint/ban-ts-comment`
- `@typescript-eslint/explicit-member-accessibility`
- `@typescript-eslint/no-empty-function`
- `@typescript-eslint/no-non-null-asserted-optional-chain`
- `@typescript-eslint/no-unused-vars`

**Fixed issue: [IEX-2405: Fix TypeScript unused vars & misc lint violations](https://youtrack.is.adyen.com/issue/IEX-2405)**


src/components/internal/Calendar/calendar/facade/Calendar.ts
```diff
@@ -634,13 +634,13 @@ export default class Calendar {
         switch (this.#highlightSelection) {
             case SELECT_MANY:
                 if (!boolOrTrue(this.#frame?.blankSelection) && this.#rangeOffsets) {
-                    this.#rangeHighlight(this.#frame?.selectionStart!, SELECTION_TO, this.#rangeOffsets);
+                    this.#rangeHighlight(this.#frame!.selectionStart!, SELECTION_TO, this.#rangeOffsets);
```
samuelsa-adyen, line 661: Is this null assertion better than keeping the safeguard of the optional chaining?

src/components/internal/Calendar/calendar/facade/Calendar.ts
```diff
@@ -634,13 +634,13 @@ export default class Calendar {
...
                 }
                 break;
             case SELECT_ONE:
                 if (!boolOrTrue(this.#frame?.blankSelection)) {
                     const restamper = withTimezone(this.#frame?.timezone);
-                    const restampedDate = new Date(timezoneToSystem(restamper, this.#frame?.selectionStart!));
+                    const restampedDate = new Date(timezoneToSystem(restamper, this.#frame!.selectionStart!));
```
samuelsa-adyen, line 667: Same here

src/components/internal/Calendar/calendar/facade/Calendar.ts
glad-adyen: If `this.#frame` is nullish (`null`/`undefined`), this block is never entered (due to the result of `boolOrTrue`). TypeScript cannot infer this narrowing from `boolOrTrue`'s return type, hence the `!` assertions, but at runtime `this.#frame` is guaranteed non-null inside this block.

Likewise, `selectionStart` is always set when `blankSelection` is `false` (a non-blank selection always has a defined start point).

These are intentional assertions, not unsafe shortcuts.

src/components/internal/Calendar/calendar/facade/Calendar.ts
glad-adyen: Ditto: previous [comment](https://github.com/Adyen/adyen-platform-experience-web/pull/1082#discussion_r3039039485)

samuelsa-adyen: Maybe we could add a safeguard to the conditional then?

something like 
```
                    const frame = this.#frame;
                    if (!boolOrTrue(frame?.blankSelection) && frame?.selectionStart) {
                        const restamper = withTimezone(frame.timezone);
                        const restampedDate = new Date(timezoneToSystem(restamper, frame.selectionStart));
                        frame.updateSelection(systemToTimezone(restamper, restampedDate.setHours(23, 59, 59, 999)), SELECTION_TO);
                    }
```

glad-adyen: This is essentially the same as the previous, only with an initial `frame` const assignment. Maybe I am missing something — how would the safeguard  you're proposing here work?

samuelsa-adyen: There are no no-null assertions, instead, the operation does not execute if `frame?.selectionStart` is not defined.

samuelsa-adyen (approved):

german664 (approved):

PR #1100, opened 2026-04-14
Extract shared SCSS into shared/style

german664:
Move SCSS tokens, mixins, variables, and breakpoints from `src/style/ into packages/shared/style/` as `@integration-components/style`. Rewrite all imports to use the new package path.


packages/shared/style/base.scss
samuelsa-adyen, line 1: What do you think about renaming this into `base.scss` since the folder is already called `shared`? Also, we could include this import into `index.scss` and simplify

src/components/shared.scss
samuelsa-adyen, line 1: Which means we could remove this file

packages/shared/style/base.scss
german664: Good call! ✅

src/components/shared.scss
german664: ✅

samuelsa-adyen (dismissed):

samuelsa-adyen (approved):

PR #1116, opened 2026-04-22
fix: random uuid generation bug

glad-adyen:
## Summary
This PR introduces changes to fix a bug with the random UUID_4 generator util function (`uuid()`), ensuring that the bytes are read correctly.


isuert (commented):
I'd suggest expanding unit test coverage to make sure we have no regression if we ever change this implementation.

glad-adyen:
> I'd suggest expanding unit test coverage to make sure we have no regression if we ever change this implementation.

It already has unit tests that exercise the part of the behavior that can be tested. The details of the new changes are however internal implementation details.

isuert (approved):

german664 (approved):

PR #1136, opened 2026-04-28
Fix tests

german664:


glad-adyen (approved):

packages/shared/testing/src/playwright/session-request-function.ts
```diff
@@ -1,5 +1,5 @@
 import { test as base, APIRequestContext, request } from '@playwright/test';
-import { resolveEnvironment } from '@integration-components/core';
+import { resolveEnvironment } from '@integration-components/core/utils';
```
samuelsa-adyen, line 2: Could we consider switching the declarations that are causing this issue to interfaces for the classes? This way we could still import from the barrel and prevent this from coming up in the future.

```
export interface SetupContext {
    loadingContext?: Core<any>['loadingContext'];
    analyticsPayload?: Array<URLSearchParams> | undefined;
    analyticsEnabled?: boolean;
    refresh: (signal: AbortSignal) => Promise<void>;
}
```

```
export interface AuthSessionSpecification {
    errorHandler: onErrorHandler | null;
    readonly autoRefresh: _AuthSessionSpecification['autoRefresh'];
    readonly onRefresh: _AuthSessionSpecification['onRefresh'];
}
```

ElifDogaliBal (approved):

packages/shared/testing/src/playwright/session-request-function.ts
german664: I am planning to do this in a separate PR, focused on a more robust fix. But for now I think this is good enough. Wdyt?

samuelsa-adyen: Seems reasonable ✅

PR #1138, opened 2026-04-29
Extract domain: payouts

german664:
Extracts the Payouts domain into `packages/domains/payouts/`

**Fixed issue**:  [IEX-2474](https://youtrack.is.adyen.com/issue/IEX-2474)


mocks/mock-data/index.ts
```diff
@@ -2,6 +2,6 @@ export * from './balanceAccounts';
 export * from './capital';
 export * from './disputes';
 export * from './payByLink';
-export * from './payouts';
+export * from '../../packages/domains/payouts/mocks/mock-data/payouts';
```
ElifDogaliBal: I think this export not needed since it is imported from packages

packages/domains/payouts/preact/src/internal/DataOverviewDetails/DataOverviewDetails.tsx
ElifDogaliBal, line 1: Should this file located under payouts domain or in a shared folder?

mocks/mock-data/index.ts
german664: ✅

packages/domains/payouts/preact/src/internal/DataOverviewDetails/DataOverviewDetails.tsx
german664: This is a version specifically for payouts. I think we can consider to have a shared component (like we used to) but it would be out of scope of this PR.

ElifDogaliBal (dismissed):

packages/domains/payouts/domain/src/PayoutsOverview/types.ts
```diff
@@ -17,7 +17,7 @@ export interface PayoutsOverviewProps extends UIElementProps {
     onRecordSelection?: (selection: { balanceAccountId: string; date: string; showModal: () => void }) => any;
     dataCustomization?: {
         list?: PayoutsListCustomization;
-        details?: never;
+        details?: PayoutDetailsCustomization;
```
glad-adyen, line 20: This change is not 1:1. Do you remember why we had `never` here?

src/components/internal/DataOverviewDetails/DataOverviewDetailsSkeleton.scss
glad-adyen, line 1: This is a new file. Why was it introduced?

package.json
```diff
@@ -76,8 +76,8 @@
     "test:integration:debug": "cross-env PWDEBUG=console TEST_ENV=1 NODE_OPTIONS='--no-experimental-strip-types' playwright test --debug --project local-chrome",
     "test:integration:ui": "npm run test:integration -- --ui",
     "translations:sort": "bash scripts/process-translations/sort-translations-json.sh",
-    "types:check": "tsc --preserveWatchOutput && tsc --noEmit --project packages/domains/reports/tsconfig.json",
-    "types:watch": "tsc --preserveWatchOutput --watch && tsc --noEmit --project packages/domains/reports/tsconfig.json --watch && tsc --project tsconfig-build.json --watch",
+    "types:check": "tsc --preserveWatchOutput && tsc --noEmit --project packages/domains/reports/tsconfig.json && tsc --noEmit --project packages/domains/payouts/tsconfig.json",
+    "types:watch": "tsc --preserveWatchOutput --watch && tsc --noEmit --project packages/domains/reports/tsconfig.json --watch && tsc --noEmit --project packages/domains/payouts/tsconfig.json --watch && tsc --project tsconfig-build.json --watch",
     "version": "./scripts/bump-version.sh"
```
glad-adyen: What about moving these to scripts defined in each package's `project.json` and then orchestrate with NX?

german664: ✅

src/components/internal/DataOverviewDetails/DataOverviewDetailsSkeleton.scss
german664: I tried to split the styles for the skeleton (which is used in different domains) from the one in `DataOverviewDetails.scss` (that is used in payouts). In a next PR we can decide how to deal with `DataOverviewDetailsSkeleton.scss`. I'm thinking maybe moving it to `@integration-components/style`, but still undecided

glad-adyen (approved):

ElifDogaliBal (approved):

PR #1139, opened 2026-04-29
Extract CoreBase for cross-framework use

samuelsa-adyen:
## Summary
- Add packages/shared/core/src/CoreBase.ts and re-export from the package's
      index.
- Convert src/core/core.ts to extend CoreBase; remove the now-duplicated
  loadingContext/session/cdn fetcher initialisation and replace setOptions()
  with the protected onOptionsChanged() hook from the base class.
- Mark vue and vue-i18n as optional peer dependencies of the core package
  so a future Vue subpath can import them without forcing Preact-only
  consumers to install them.
- Include shims-vue.d.ts so .vue imports type-check in the shared core
  tsconfig.

## Tested scenarios

**Fixed issue**:


packages/shared/core/src/CoreBase.ts
```diff
@@ -0,0 +1,104 @@
...
+    protected constructor(options: O) {
+        this.options = { environment: FALLBACK_ENV, ...options } as O;
+        const { apiUrl, cdnAssetsUrl, cdnConfigUrl } = this.resolveEnvironment();
+
+        this.loadingContext = options.loadingContext || process.env.VITE_APP_LOADING_CONTEXT || apiUrl;
+        this.analyticsEnabled = options.analytics?.enabled ?? true;
+        this.session.analyticsEnabled = this.analyticsEnabled;
+        this.getCdnConfig = getConfigFromCdn({ url: cdnConfigUrl });
+        this.getCdnDataset = getDatasetFromCdn({ url: `${cdnAssetsUrl}/datasets` });
+    }
```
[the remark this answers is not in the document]
glad-adyen: This is an important point.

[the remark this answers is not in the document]
samuelsa-adyen: Bottom line of the constructor now calls ` this.setOptions(this.options);` which defines `loadingContext` and `onSessionCreate`

src/core/core.ts
```diff
@@ -105,22 +96,13 @@ class Core<AvailableTranslations extends TranslationSourceRecord[] = [], CustomT
...
+        if (!this.localization) return;
         this.localization.locale = this.options?.locale;
         this.localization.customTranslations = this.options?.translations;
-
-        this.session.loadingContext = this.loadingContext;
-        this.session.onSessionCreate = this.options.onSessionCreate;
-
-        return this;
-    };
+    }
```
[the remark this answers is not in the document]
samuelsa-adyen: Done, added the extra check for environment updates.

packages/shared/core/src/Core.ts
```diff
@@ -0,0 +1,104 @@
...
+    protected setOptions(options: Partial<O>): this {
+        this.options = { ...this.options, ...options };
+
+        this.onOptionsChanged(options);
+
+        this.session.loadingContext = this.loadingContext;
+        this.session.onSessionCreate = this.options.onSessionCreate;
+
+        return this;
+    }
```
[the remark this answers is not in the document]
samuelsa-adyen: added extra check for loadingContext and analytics changes

packages/shared/core/src/CoreBase.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    /**
+     * Merge incoming options, propagate locale / custom translations to the shared
+     * `Localization`, hand off to the subclass hook, then sync the session.
+     */
+    protected setOptions(options: Partial<O>): this {
+        const environmentChanged = options.environment !== undefined && options.environment !== this.options.environment;
+        const loadingContextChanged = options.loadingContext !== undefined && options.loadingContext !== this.options.loadingContext;
+        const analyticsChanged = options.analytics !== undefined;
```
glad-adyen: Why omit the sameness check for `analyticsChanged` here? If `options.analytics` is the same value as `this.options.analytics`, the check will always return `true`, suggesting that `analyticsChanged`.

samuelsa-adyen: True, added the extra check.

german664 (dismissed):

glad-adyen (commented):
Looking at this PR again, I don't see the need of an abstraction for `Core`, since it is already framework-agnostic.

I believe the framework distinction is best suited for the `BaseElement` class (which is responsible for framework-specific rendering, mounting, unmounting, etc.).

packages/shared/core/src/preact/BaseElement.ts
```diff
@@ -13,6 +13,10 @@ class BaseElement<P> {
     public props: P & BaseElementProps;
     public state: BaseElementState = {};
 
+    public get core(): any {
+        return this.props.core;
+    }
```
glad-adyen, line 18: Why do we need a `core` get accessor? `this.props.core` is already publicly accessible.

packages/shared/core/src/Core.ts
```diff
@@ -0,0 +1,210 @@
...
+        return this.localization.i18n;
+    }
+
+    /**
+     * Optional subclass hook for framework-specific reactions to option changes
+     * (e.g. rebuilding env-bound assets when `options.environment` changes).
+     * Called from `setOptions` *after* `this.options` has been merged and shared
+     * localization state has been refreshed.
+     */
+    protected onOptionsChanged?(options: Partial<O>): void;
```
glad-adyen: This becomes redundant

packages/shared/core/src/preact/BaseElement.ts
samuelsa-adyen: This is because the Vue implementation does not read directly from props.core as the preact one does, so we need the getter to abstract this difference.

packages/shared/core/src/Core.ts
samuelsa-adyen: Good catch, removed ✅

packages/shared/core/src/Core.ts
```diff
@@ -0,0 +1,200 @@
...
+export interface CoreOptionsBase {
+    environment?: DevEnvironment;
+    locale?: string;
+    onSessionCreate: SessionRequest;
+    onError?: onErrorHandler;
+    analytics?: AnalyticsConfig;
+    loadingContext?: string;
+    availableTranslations?: TranslationSourceRecord[];
+    translations?: Translations;
+}
```
glad-adyen: Not sure we would need this anymore. `CoreOptions` should be sufficient for now.

packages/shared/core/src/Core.ts
```diff
@@ -0,0 +1,200 @@
...
+
+export interface ResolvedEnvironment {
+    apiUrl: string;
+    cdnTranslationsUrl: string;
+    cdnAssetsUrl: string;
+    cdnConfigUrl: string;
+}
```
glad-adyen: We can export this `ResolvedEnvironment` interface from the same module as the `resolveEnvironment` util to keep them co-located.

packages/shared/core/src/Core.ts
```diff
@@ -0,0 +1,200 @@
...
+/**
+ * Minimal contract that framework-specific element classes (Preact BaseElement,
+ * Vue UIElement, ...) must satisfy so Core can manage them uniformly.
+ */
+export interface ManagedElement {
+    readonly _id: string;
+    readonly core: unknown;
+    update(props: any): any;
+    unmount(): any;
+}
```
glad-adyen, line 20: Do we still need this?

samuelsa-adyen: Yes, we need it to generalize the component type for both frameworks

packages/shared/core/src/Core.ts
```diff
@@ -0,0 +1,200 @@
...
+
+/**
+ * Framework-agnostic source of truth for the Core runtime. Owns option resolution,
+ * environment, session wiring, the shared `Localization` instance, asset getters, CDN
+ * helpers, the component registry, and the generic `initialize()` / `update()` lifecycle.
+ *
+ * Framework-specific rendering / mounting / unmounting lives in the element classes
+ * (Preact `BaseElement`, Vue `UIElement`, ...), not here.
+ */
+export class Core<O extends CoreOptionsBase = CoreOptionsBase> {
```
glad-adyen: The new generic types for the `Core` class deviate from the existing types. Since the user-facing `AdyenPlatformExperience()` function mirrors these same generic types, it would be safer to revert to the existing types, to avoid any potential breaking changes.

packages/shared/core/src/Core.ts
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        return this;
+    }
+
+    /**
+     * Apply a partial options patch, re-initialize, and propagate the update to
+     * every registered component that belongs to this Core instance.
+     */
+    public async update(options: Partial<O> = EMPTY_OBJECT as Partial<O>): Promise<this> {
```
glad-adyen: This used to be an instance method, but has now been changed to a class method. It shouldn't make a big difference ordinarily. But for consistency, I'd suggest changing the other public methods (e.g. `remove`, `registerComponent`) to class methods.

packages/shared/core/src/Core.ts
```diff
@@ -0,0 +1,200 @@
...
+            this.getDatasetAsset = new Assets(cdnAssetsUrl).getAsset({ extension: 'json', mainFolder: 'datasets' });
+        } else if (loadingContextChanged) {
+            const { apiUrl } = this.resolveEnvironment();
+            this.loadingContext = this.options.loadingContext || process.env.VITE_APP_LOADING_CONTEXT || apiUrl;
+        }
+
+        if (analyticsChanged) {
+            this.analyticsEnabled = this.options.analytics?.enabled ?? true;
+            this.session.analyticsEnabled = this.analyticsEnabled;
+        }
```
glad-adyen: Some of these operations are repeated between the `constructor()` and the `SetOptions()` method. A possible improvement could be to add some private helper methods.

packages/shared/core/src/Core.ts
samuelsa-adyen: Done

packages/shared/core/src/Core.ts
samuelsa-adyen: Moved to `packages/shared/core/src/types.ts`

packages/shared/core/src/Core.ts
samuelsa-adyen: Rolled back to previous types.

packages/shared/core/src/Core.ts
samuelsa-adyen: Done, made them class methods

packages/shared/core/src/Core.ts
samuelsa-adyen: Done, created two auxiliary methods, `applyAnalyticsOptions` and `applyEnvironmentAssets`

glad-adyen (approved):

german664 (approved):

PR #1148, opened 2026-05-04
Feature/vue platform infrastructure

samuelsa-adyen:
## Tested scenarios

**Fixed issue**:


packages/tools/storybook/src/shared/styles.scss
```diff
@@ -13,6 +13,7 @@ body {
     text-size-adjust: 100%;
 }
 
+/* stylelint-disable selector-class-pattern */
```
glad-adyen: Why do you need this disable comment? Isn't there already an exception in the storybook package's `.stylelintrc.cjs` file?

packages/tools/storybook/src/shared/previewDefaults.ts
```diff
@@ -1,7 +1,25 @@
...
+setup(app => {
+    app.use(
+        createI18n({
+            legacy: false,
+            locale: 'en-US',
+            fallbackLocale: 'en-US',
+            messages: { 'en-US': {} },
+        })
+    );
+});
```
glad-adyen: This is Vue-specific. Why is it in the shared `previewDefaults.ts` file instead of the Vue-specific `preview.ts` file?

packages/tools/storybook/src/.storybook/vue/styles.scss
```diff
@@ -0,0 +1,28 @@
...
+    justify-content: center;
+    width: 100%;
+}
+
+.compact-component-wrapper {
+    margin: auto;
+    max-width: 600px;
+    width: 100%;
+}
+/* stylelint-enable selector-class-pattern */
```
glad-adyen: Why do we need another `styles.scss` for the Vue storybook container? Are we not able to reuse the shared styles? If overrides are needed, then we should only add overrides to the framework-specific SCSS files.

packages/tools/storybook/src/.storybook/vue/index.ts
```diff
@@ -0,0 +1,3 @@
+export { default as Container } from './Container.vue';
+export * from './controls';
+export * from './constants';
```
glad-adyen: Do we really need an index file here?

packages/tools/storybook/src/.storybook/vue/main.ts
```diff
@@ -1,17 +1,18 @@
...
 
 const root = '../../../../../..';
 const rootDir = resolve(import.meta.dirname, root);
 
 const config: StorybookConfig = {
-    // Framework-scoped under {domain}/vue
-    // The framework split is at the top level of each domain.
+    // Framework-scoped under {domain}/vue/ — the preact/vue split is at the top level
+    // of each domain. Phase 1 ships a single placeholder smoke under
+    // packages/domains/reports/vue/stories/. Real domain Vue stories land in 04-03.
```
glad-adyen: Why override the comments here? I think the previous form is more concise and appropriate (without introducing ambiguous context).

packages/tools/storybook/src/.storybook/vue/controls.ts
glad-adyen, line 1: Can this be moved to shared folder instead, and shared between Preact and Vue?

packages/tools/storybook/src/.storybook/vue/constants.ts
glad-adyen, line 1: Can this be moved to shared folder, and shared between Preact and Vue?

packages/shared/core/src/vue/Context/eventDispatcher/EventDispatcherProvider.vue
```diff
@@ -0,0 +1,98 @@
...
+        unsubscribe: (callback: (event: EventQueueItem) => void) => {
+            subscribers.delete(callback);
+        },
+        updateBaseTrackingPayload: (payload: Record<string, any>) => {
+            basePayload = { ...basePayload, ...payload };
+        },
+        updateSharedEventProperties: (properties: Record<string, any>) => {
+            sharedProperties = { ...sharedProperties, ...properties };
+        },
+    };
```
glad-adyen: Why are we doing this, instead of just constructing a new instance of `UserEvents`?

packages/shared/testing/src/storybook-helpers/sessionRequest.ts
glad-adyen, line 1: This util is framework-agnostic. Do we need another copy of it?

.eslintrc.cjs
```diff
@@ -76,6 +76,8 @@ module.exports = {
                     'packages/domains/*/tests/**/*.{ts,tsx}',
                     'packages/domains/*/**/stories/**/*.{ts,tsx}',
                     'packages/domains/*/mocks/**/*.{ts,tsx}',
+                    // Storybook tooling packages — everything in packages/tools/ is dev-only.
+                    'packages/tools/**/*.{ts,tsx,vue}',
```
glad-adyen: Check this config against the storybook package's `.eslintrc.cjs` file.

packages/shared/core/src/vue/UIElement.scss
glad-adyen, line 1: These styles are the same between Preact and Vue. How can they be reused rather than duplicated?

samuelsa-adyen: Duplication for this one is intended. Coupling both implementations that much together would remove some of the flexibility that we have for these components.

glad-adyen: I wouldn't say the implementations are/should be coupled though. But out of curiosity, what limitations have been observed so far?

samuelsa-adyen: For instance, the width of the capital components for instance varies on the designs, since they're not final, we have no way of knowing if they will stay `600px`, classes and the applications for box-sizing might change due to the bento components being used, etc.

glad-adyen: Fair point. What do you think about (S)CSS overrides?

packages/shared/testing/src/storybook-helpers/sessionRequest.ts
samuelsa-adyen: I was trying to make it `ts` instead of moving, I think it probably got relocated during the storybook refactoring. Moved to a single copy of `packages/shared/testing/src/storybook-helpers/sessionRequest.ts`

packages/tools/storybook/src/.storybook/vue/constants.ts
samuelsa-adyen: Deleted, we're exporting this functionality from somewhere else.

packages/tools/storybook/src/.storybook/vue/controls.ts
samuelsa-adyen: Deleted. Can be imported from `packages/shared/testing/src/storybook-helpers/controls.ts`

packages/tools/storybook/src/.storybook/vue/main.ts
samuelsa-adyen: Not sure if we even need the comment, seems kind of redundant. I removed it, lmk if you think it could be useful.

packages/tools/storybook/src/shared/previewDefaults.ts
samuelsa-adyen: Moved to `packages/tools/storybook/src/.storybook/vue/preview.ts`

packages/tools/storybook/src/shared/styles.scss
samuelsa-adyen: Removed. It was added before the refactoring of the storybook package.

packages/tools/storybook/src/.storybook/vue/index.ts
samuelsa-adyen: Not really - removed.

packages/tools/storybook/src/.storybook/vue/styles.scss
samuelsa-adyen: I was thinking of flexibility too for this one, but since it's storybook-specific, I think we can standardize. Removed.

packages/shared/core/src/vue/Context/eventDispatcher/EventDispatcherProvider.vue
samuelsa-adyen: Refactored, now using `packages/shared/core/src/EventDispatcher/eventDispatcher/user-events.ts` which uses UserEvents

.eslintrc.cjs
samuelsa-adyen: Not needed anymore - removed.

packages/shared/core/src/vue/UIElement.scss
samuelsa-adyen: We'd have to increase specificity for the selector, and might be inconvenient for the final users.

glad-adyen: What I meant was something like:

```scss
// import shared styles
@use 'shared';

// override style
.adyen-pe-component {/* ... */}
```

samuelsa-adyen: Added the `packages/shared/core/src/UIElementBase.scss` file.

packages/shared/composables-vue/src/useCustomColumnsData.ts
```diff
@@ -0,0 +1,67 @@
...
+        } else {
+            customRecords.value = currentRecords;
+        }
+    }
+
+    watch(
+        [() => records(), () => hasCustomColumn?.(), () => onDataRetrieve?.()],
+        () => {
+            void recompute();
+        },
```
glad-adyen: **Not so important:**
Since this returns `void`, it can be written like so:

```suggestion
        () => void recompute(),
```

packages/shared/core/src/vue/Context/eventDispatcher/useEventDispatcherContext.ts
```diff
@@ -0,0 +1,27 @@
...
+
+    return new Proxy({} as Partial<UserEvents>, {
+        get(_target, prop: string) {
+            const current = (injected.value ?? {}) as Record<string, unknown>;
+            return current[prop];
+        },
+        has(_target, prop: string) {
+            return prop in (injected.value ?? {});
+        },
+    });
```
glad-adyen: Is there any situation where `injected` is not a **ref**? It seems like it is created using `computed()`, which makes the value technically a ref.

packages/shared/core/src/vue/UIElement.ts
```diff
@@ -0,0 +1,96 @@
...
+     */
+    public get core(): any {
+        return (this._props as any)?.core;
+    }
+
+    constructor(component: Component, props: Props, componentName?: string) {
+        this._component = component;
+        this._props = reactive({ ...(props as Record<string, unknown>) }) as Props;
+        this._componentName = componentName;
+    }
```
glad-adyen, line 51: There are a few things missing from this implementation, such as component instance registration (`this.props.core.registerComponent(this)`) that happens in the constructor of `BaseElement` for Preact component.

Since this `UIElement` class (for external Vue components) aims to fill the role of `UIElement` < `BaseElement` for external Preact components, I'd recommend double-checking to fill the necessary gaps.

packages/shared/core/src/vue/Context/eventDispatcher/useEventDispatcherContext.ts
samuelsa-adyen: True, removed the extra conditional

packages/shared/composables-vue/src/useCustomColumnsData.ts
samuelsa-adyen: simplified ✅

packages/shared/core/src/vue/UIElement.ts
samuelsa-adyen: Added: 
1. `core.registerComponent(this)` to the constructor
2. `remove()` method 
3. `type` getter
4. `displayName` getter
5. `customClassNames` field

For parity with UIElement + BaseElement from Preact.

packages/shared/core/src/vue/Context/eventDispatcher/useEventDispatcherContext.ts
```diff
@@ -0,0 +1,23 @@
...
```
glad-adyen, line 22: I imagine you are trying to avoid noisy `.value` access of the ref's value (unwrapping the ref) downstream. However, returning a proxy here might harm the reactivity of the ref downstream. Two reasonable options:

1. Wrap the ref in `reactive()` to preserve reactivity of the ref downstream without needing `.value` access. The only downside to this is that destructuring from the result will also break reactivity.

```ts
    return reactive(injected);
```

2. Introduce reactivity into the proxy `get` and `has` traps (using `unref()`) so that it is propagated downstream. Now destructuring should also work, however, destructured primitives will loose their reactivity.

```ts
    return new Proxy({} as Partial<UserEvents>, {
        get(_target, prop: string) {
            const current = unref(injected) as Record<string, any>;
            return current?.[prop];
        },
        has(_target, prop: string) {
            return prop in (unref(injected) ?? {});
        },
    });
```

samuelsa-adyen: Good call. Went with option 2. Using unref() in both proxy traps so reactivity is properly propagated downstream. Also added optional chaining in the get trap to handle the null/undefined case more safely.

glad-adyen (dismissed):

glad-adyen (approved):

german664 (approved):

PR #1149, opened 2026-05-04
Implement Reports overview component in Vue

samuelsa-adyen:
## Tested scenarios

**Fixed issue**:


packages/domains/reports/vue/src/ReportsOverview/components/ReportsTable.vue
```diff
@@ -0,0 +1,254 @@
...
+        };
+    } else {
+        alert.value = {
+            title: i18n.get('reports.overview.errors.download'),
+            description: i18n.get('reports.overview.errors.retryDownload'),
+        };
+    }
+}
+
+function getReportType(type: string): string {
```
german664: I think we can replace this with the function we have in `packages/domains/reports/domain/src/ReportsOverview/translations.ts`

So we could do `import { getReportType } from '@integration-components/reports/domain';`

And use it like 

`getReportType(i18n, item.type) ?? item.type`

packages/domains/reports/vue/src/ReportsOverview/types.ts
```diff
@@ -0,0 +1,38 @@
...
+import type { StringWithAutocompleteOptions } from '@integration-components/utils/types';
+
+export interface DataCustomizationList<Fields extends string> {
+    fields?: CustomColumn<StringWithAutocompleteOptions<Fields>>[];
+    onDataRetrieve?: OnDataRetrievedCallback<IReport[]>;
+}
+
+// ── Component prop types ──
+
+export interface ReportsOverviewExternalProps {
```
german664: `ReportsOverviewExternalProps` is already defined in `@integration-components/reports/domain`. The Vue specific addition is only `core: CoreInstance`. So I think we could do something like:

```
     import type { ReportsOverviewProps } from '@integration-components/reports/domain';
     import type { CoreInstance } from '@integration-components/core/vue';

     export interface ReportsOverviewExternalProps extends Omit<ReportsOverviewProps, 'onContactSupport' | 'hideTitle'> {
         core: CoreInstance;
         // add any other Vue only props here
     }
```

packages/domains/reports/vue/src/ReportsOverview/constants.ts
german664, line 1: Some of this constants are also in `packages/domains/reports/domain/src/ReportsOverview/constants.ts`.

What do you think of unifying them?

packages/domains/reports/vue/src/ReportsOverview/styles/ReportsOverview.scss
```diff
@@ -0,0 +1,24 @@
+@use '@integration-components/style' as style;
+
+.adyen-pe-reports-overview {
+    display: flex;
+    flex-direction: column;
+    gap: 16px;
```
german664: Should we use `style.token(spacer-*)` in this file instead of the absolute value?

packages/tools/storybook/src/index.ts
glad-adyen, line 1: Do we really need this file?

packages/tools/storybook/src/.storybook/vue/index.ts
glad-adyen, line 1: Do we really need this file?

packages/tools/storybook/src/index.ts
samuelsa-adyen: Yes. We need to export the `Container` component to use it in [here](https://github.com/Adyen/adyen-platform-experience-web/pull/1149/changes#diff-967edf65a9fc5fceffdd0c0b11421559d57a0aa2761b91ceeaaf8b24d7576330R5)

packages/tools/storybook/src/.storybook/vue/index.ts
samuelsa-adyen: Same as the last comment, just want to make the exports cleaner.

packages/tools/storybook/src/index.ts
glad-adyen: I don't think we need to export it from the storybook package. It can be defined in the `preview.ts` file of the Vue storybook setup without leaking out from the storybook package. For context, see the way it is setup for Preact's `preview.tsx`.

In the `setup` call in the Vue storybook's `preview.ts`, you can also register a global component. Something like:
```ts
    app.component('story-container', Container);
```

And then you should be able to use component in the stories without exporting 🤞

packages/tools/storybook/src/.storybook/vue/index.ts
glad-adyen: https://github.com/Adyen/adyen-platform-experience-web/pull/1149#discussion_r3234498317

packages/domains/reports/vue/src/ReportsOverview/components/ReportsTable.vue
samuelsa-adyen: Since they're on the same domain, I added relative paths, but now we're using the same function

packages/domains/reports/vue/src/ReportsOverview/types.ts
samuelsa-adyen: Done!

packages/domains/reports/vue/src/ReportsOverview/constants.ts
samuelsa-adyen: Sure, I deleted the `Vue` specific file

packages/domains/reports/vue/src/ReportsOverview/styles/ReportsOverview.scss
samuelsa-adyen: Nice catch, done ✅

german664 (approved):

glad-adyen (approved):

PR #1150, opened 2026-05-04
Enable multiple-version resource generation

isuert:
This PR improves the script for generating schema. Now the script no longer targets a single API version from `package.json`, but  fetches all versioned files per resource folder (skipping `*-latest.*` files), and generates a .ts file per version.

On top of this the PR fixed the resources path in the schema generation script, improves the code of the later, changes the naming convention for V1 resources, and removes unnecessary resource files.


src/types/api/resources/BalanceAccountsResource.ts
glad-adyen, line 1: Why deleting these files in the `src/types/api/resources`?

I think they are still serving some purpose in the migration process, to keep the SDK package working as expected. I would recommend double-checking with @german664 before deleting them.

Also, I am thinking deleting them might be out of scope for this PR.

isuert: They are not necessary, I double checked with @german664.

glad-adyen (dismissed):

glad-adyen (approved):

mocks/mock-server/capital.ts
```diff
@@ -22,10 +22,10 @@ import { endpoints } from '../../endpoints/endpoints';
 import { DefaultBodyType, http, HttpResponse, StrictRequest } from 'msw';
 import { calculateGrant } from './utils/utils';
 import { delay, getHandlerCallback, mocksFactory } from '@integration-components/testing/msw';
-import { paths as CapitalGrantOfferPaths } from '../../src/types/api/resources/CapitalGrantOffersResource';
-import { paths as CapitalGrantsPaths } from '../../src/types/api/resources/CapitalGrantsResource';
-import { paths as CapitalMissingActionsPaths } from '../../src/types/api/resources/CapitalMissingActionsResource';
-import { paths as OnboardingSessionPaths } from '../../src/types/api/resources/OnboardingConfigurationResource';
+import { paths as capitalGrantOffersPaths } from '../../packages/shared/types/src/api/resources/CapitalGrantOffersResourceV1';
```
german664, line 25: Why changing the name of the paths in this file? Like `capitalGrantOffersPaths` for example

packages/shared/types/src/api/models/capital.ts
german664, line 1: I wonder why we are not using the capital first letter for the components types in this file. I think it's a pretty common convention to do so

german664 (approved):

PR #1151, opened 2026-05-05
Setup layered defines config

glad-adyen:
## Summary
This PR introduces changes to setup layers for Vite's `define` config across package boundaries. The `base-env` config can be used by any package without package dependency requirements. The `build-env` config, on the other hand, has dependency on the shared utils package.

Some minor changes include:
- Rearranging the line order of imports and declarations
- Rearranging the config variables in alphabetical order


config/defines/base-env.ts
```diff
@@ -1,31 +1,28 @@
...
         'process.env.SESSION_MAX_AGE_MS': JSON.stringify(isDevelopment ? session.maxAgeMs || null : undefined),
         'process.env.SESSION_PERMISSIONS': JSON.stringify(session.permissions || null),
         'process.env.TEST_ENV': JSON.stringify(process.env.TEST_ENV),
         'process.env.USE_CDN': JSON.stringify(app.useCdn ?? null),
+        'process.env.VITE_APP_LOADING_CONTEXT': JSON.stringify(isDevelopment ? app.loadingContext || null : null),
+        'process.env.VITE_APP_PORT': JSON.stringify(app.port || null),
+        'process.env.VITE_APP_URL': JSON.stringify(process.env.DEPLOY_PRIME_URL?.replace('main--', '') || app.url || null),
+        'process.env.VITE_LOCAL_ASSETS': JSON.stringify(localAssets),
+        'process.env.VITE_MODE': JSON.stringify(process.env.VITE_MODE ?? mode),
         'process.env.VITE_TEST_CDN_ASSETS': JSON.stringify(testCdnAssets),
```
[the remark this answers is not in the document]
glad-adyen: Done ✅

vite.config.ts
```diff
@@ -4,9 +4,9 @@ import { resolve } from 'node:path';
 import { preact } from '@preact/preset-vite';
 import { visualizer } from 'rollup-plugin-visualizer';
 import { realApiProxies } from './endpoints/realApiProxies';
+import { getBuildEnvDefines } from './config/defines/build-env';
 import { getEnvironment } from './envs/getEnvs';
 import packageJson from './package.json';
```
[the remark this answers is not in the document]
glad-adyen: Out of scope, ignoring for now

config/defines/base-env.ts
```diff
@@ -1,31 +1,28 @@
...
-        'process.env.VITE_APP_PORT': JSON.stringify(app.port || null),
-        'process.env.VITE_APP_URL': JSON.stringify(process.env.DEPLOY_PRIME_URL?.replace('main--', '') || app.url || null),
-        'process.env.VITE_APP_LOADING_CONTEXT': JSON.stringify(isDevelopment ? app.loadingContext || null : null),
-        'process.env.VITE_LOCAL_ASSETS': JSON.stringify(localAssets),
-        'process.env.VITE_BUILD_ID': JSON.stringify(ADYEN_BUILD_ID),
-        'process.env.VITE_COMMIT_BRANCH': JSON.stringify(COMMIT_BRANCH),
-        'process.env.VITE_COMMIT_HASH': JSON.stringify(COMMIT_HASH),
-        'process.env.VITE_MODE': JSON.stringify(process.env.VITE_MODE ?? mode),
-        'process.env.SDK_VERSION': JSON.stringify(ADYEN_FP_VERSION),
+        'process.env.SDK_VERSION': JSON.stringify(packageJson.version),
```
samuelsa-adyen: If we are using the `version()` script in `config/defines/build-env.ts`, could we perhaps standardize and get the package.json also from the same script? seems like it is also what we were doing before

samuelsa-adyen (dismissed):

config/defines/base-env.ts
glad-adyen: This is exactly the point of breaking the `define` variables. Importing `version` will require the shared utility package as dependency, which seems unnecessary for a package (e.g. storybook tooling) that only needs the base `define` variables, without the extra build variables.

samuelsa-adyen: Yes. What I mean is that in `config/defines/build-env.ts` we have `const { ADYEN_BUILD_ID, COMMIT_BRANCH, COMMIT_HASH } = version();` imported from `../version`, not an external package. We could do the same in `config/defines/base-env.ts`

Like this:
```
import version from '../version';
...
const { ADYEN_FP_VERSION } = version();
...
'process.env.SDK_VERSION': JSON.stringify(ADYEN_FP_VERSION),
```

for consistency

ElifDogaliBal (dismissed):

samuelsa-adyen (approved):

config/defines/base-env.ts
glad-adyen: Added a new named export instead of calling the `version()` function in the base.

german664 (approved):

PR #1164, opened 2026-05-12
Extract Pay by Link to domain

german664:


mocks/mock-data/index.ts
```diff
@@ -1,7 +1,7 @@
 export * from './balanceAccounts';
 export * from './capital';
 export * from '../../packages/domains/disputes/mocks/mock-data/disputes';
-export * from './payByLink';
+export * from '../../packages/domains/payByLink/mocks/mock-data';
```
glad-adyen: This module path looks different from the other paths for the other migrated domains. Is this expected?

tsconfig.base.json
```diff
@@ -29,7 +29,7 @@
             "@integration-components/core/*": ["packages/shared/core/src/*"],
             "@integration-components/core/preact": ["packages/shared/core/src/preact/index.ts"],
             "@integration-components/sdk-internal": ["packages/shared/lib/src/index.ts"],
-            "@integration-components/sdk-internal/*": ["packages/shared/lib/src/*"],
+            "@integration-components/sdk-internal/*": ["packages/shared/lib/src/*", "src/*"],
```
glad-adyen: Why do we need the additional top-level `src/*` path here?

vite.config.ts
```diff
@@ -103,6 +103,11 @@ export default defineConfig(({ mode }) => {
             port: app.port,
             proxy: realApiProxies(api, mode),
         },
+        resolve: {
+            alias: {
+                '@integration-components/sdk-internal': resolve(__dirname, 'src'),
+            },
+        },
```
glad-adyen: Wouldn't this conflict with the actual barrel export of the `sdk-internal` package? Why are we not allowing the this to be resolved through the package's root module directly?

packages/domains/payByLink/preact/src/PaymentLinkCreation/types.ts
```diff
@@ -1,6 +1,8 @@
-import { UIElementProps, DeepPartial } from '../../types';
-import { IPaymentLinkCreateRequest } from '../../../types/api/models/payByLink';
-import { FieldValues } from '../../../hooks/form/types';
+import { UIElementProps } from '@integration-components/types';
+
+export type DeepPartial<T> = T extends object ? { [K in keyof T]?: DeepPartial<T[K]> } : T;
```
glad-adyen: Where was this `DeepPartial` type before? Why do we have to create/export it here?

german664: ✅ good catch. It was in components/types.ts but only used in PayByLinkCreation, but it is a common type, so I moved it to shared types package.

vite.config.ts
german664: ✅

tsconfig.base.json
german664: ✅

mocks/mock-data/index.ts
german664: ✅

packages/shared/hooks-preact/package.json
```diff
@@ -11,7 +11,7 @@
         "./element/constants": "./src/element/constants.ts",
         "./element/getIntersectionObserver": "./src/element/getIntersectionObserver.ts",
         "./element/useClickOutside": "./src/element/useClickOutside.ts",
-        "./element/useDetachedRender": "./src/element/useDetachedRender.tsx",
+        "./element/useDetachedRender": "./src/element/useDetachedRender.ts",
```
glad-adyen, line 14: I think the `useDetachedRender` hook needs the `.tsx` extension since it does a bit of rendering.

packages/domains/payByLink/preact/src/PaymentLinkSettings/components/TermsAndConditions/Requirements/useTermsRequirementsConfig.test.ts
glad-adyen, line 1: Why are we removing this tests file?

packages/shared/hooks-preact/package.json
german664: The file handles rendering using `VNode` and `createPortal`, but it doesn't use JSX syntax directly. The `.tsx` extension is only needed if JSX is used.

packages/domains/payByLink/preact/src/PaymentLinkSettings/components/TermsAndConditions/Requirements/useTermsRequirementsConfig.test.ts
german664: I restored it. Thought it could be unnecessary and tried to avoid using a relative path, but will fix it in a future PR when I sort out the translation files.

packages/shared/hooks-preact/package.json
glad-adyen: Got it

glad-adyen (commented):
Still missing the `payByLink` configurations.

glad-adyen (approved):

glad-adyen (dismissed):

packages/domains/payByLink/mocks/endpoints.ts
```diff
@@ -0,0 +1,21 @@
...
+    expire: `${MSW_BASE_URL}/paybylink/paymentLinks/:id/expire`,
+    filters: `${MSW_BASE_URL}/paybylink/filters`,
+    installments: `${MSW_BASE_URL}/paybylink/installments`,
+    list: `${MSW_BASE_URL}/paybylink/paymentLinks`,
+    settings: `${MSW_BASE_URL}/paybylink/settings/:storeId`,
+    themes: `${MSW_BASE_URL}/paybylink/themes/:id`,
+} as const;
+
+// Cross-cutting endpoints shared with other domains
+export const CROSS_CUTTING_ENDPOINTS = {
```
samuelsa-adyen: If we are sharing these, should we move them to a domain that is widely accessible? Also, we could rename to `CROSS_DOMAIN_ENDPOINTS` since it is nomenclature we're currently using

samuelsa-adyen (dismissed):

packages/domains/payByLink/mocks/endpoints.ts
german664: great point ✅

samuelsa-adyen (approved):

glad-adyen (approved):

PR #1178, opened 2026-05-20
Feature/variable terms UI

isuert:
Adds support for variable repayment terms in the Capital offer flow. Users can now select a repayment plan from multiple term options before reviewing and submitting their financing request.

Changes:
- Added a TermSelector component to the offer selection screen, allowing users to pick a repayment plan with the associated daily rate displayed per option
- Reworked CapitalOfferSummary to show a financing summary (amount, fees, total repayment) alongside the term-specific details (daily rate, APR, repayment threshold, expected and maximum repayment period/date)
- Updated CapitalOfferSelection layout and styles to accommodate the term selector
- Made Card component support a clickable variant, used for selectable term cards
- Updated integration tests to reflect the new flow and updated copies


src/components/external/CapitalOffer/components/CapitalOfferSelection/CapitalOfferSelection.tsx
```diff
@@ -35,32 +36,27 @@ const sharedAnalyticsEventProperties = {
     subCategory: 'Business financing offer',
 } as const;
 
-const LoadingSkeleton = () => (
+const LoadingSkeleton = ({ hasSingleTerm }: { hasSingleTerm: boolean }) => (
```
german664, line 40: To make it more scalable, instead of relying on a single boolean value (e.g., hasSingleTerm) that might not be enough` in the future, what if we pass the desired number of rows as a prop and set a default?

So, instead of using `{ hasSingleTerm: boolean }` and `Array(hasSingleTerm ? 6 : 5)`, we could use `{ skeletonRows: number }` and `Array(skeletonRows)`.

That way, if the layout of the Capital skeleton changes again in the future due to external factors, we can just update the number instead of adding new props.

isuert: I don't think scalability is a concern here - this component is not reused anywhere, it's just a local component used to abstract away some markup from the CapitalOfferSelection, including the number of rows.

Here is how it's used:
```ts
{isLoadingIndicatorVisible ? (
  <LoadingSkeleton hasSingleTerm={hasSingleTerm} />
) : matchedOffer ? (
  <InformationDisplay data={matchedOffer} hasSingleTerm={hasSingleTerm} />
) : null}
```

src/components/internal/Card/types.ts
```diff
@@ -1,6 +1,7 @@
 import { ComponentChild } from 'preact';
 
 export interface CardProps {
+    onClick?: () => void;
     onClickHandler?: () => void;
```
glad-adyen: What's the difference between the `onClick` prop (added here) and the `onClickHandler` prop (already existing here)? If the `onClickHandler` is no longer being used, can it be removed?

src/components/internal/Card/Card.tsx
```diff
@@ -84,6 +96,7 @@ const Card = ({
                 [CARD_NO_OUTLINE]: noOutline,
                 [CARD_NO_PADDING]: noPadding,
                 [CARD_EXPANDABLE_CLASS]: expandable,
+                [CARD_CLICKABLE]: !!onClick,
```
glad-adyen, line 108: As it seems, the card is also clickable when it is expandable. Why is the `CARD_CLICKABLE` class name only introduced for _truthy_ `onClick` prop value?

src/components/internal/Card/Card.tsx
```diff
@@ -56,25 +58,35 @@ const Card = ({
...
                 case InteractionKeyCode.SPACE:
                     evt.preventDefault();
-                    toggleExpansion();
+                    if (expandable) toggleExpansion();
+                    else onClick?.();
                     return;
             }
         },
-        [toggleExpansion]
+        [expandable, onClick, toggleExpansion]
```
glad-adyen: What is the rationale for the change here? The `toggleExpansion` callback already does the `expandable` check. I reason the following should be sufficient.
```suggestion
                    toggleExpansion();
                    onClick?.();
                    return;
            }
        },
        [onClick, toggleExpansion]
```

To make it more reusable, I'd suggest defining a callback like so:
```ts
const handleClick = useCallback(() => {
    toggleExpansion();
    onClick?.();
}, [onClick, toggleExpansion]);
```

and then use it like so:
```suggestion
                    handleClick();
                    return;
            }
        },
        [handleClick]
```

src/components/internal/Card/Card.tsx
```diff
@@ -56,25 +58,35 @@ const Card = ({
...
+        if (onClick) {
+            return {
+                role: 'button' as AriaRole,
+                tabIndex: 0,
+                onClick,
+                onKeyDown,
+            };
+        }
+        return {};
+    }, [expandable, onClick, showContent, cardId, onKeyDown, toggleExpansion]);
```
glad-adyen: Following the suggestion [here](https://github.com/Adyen/adyen-platform-experience-web/pull/1178/changes#r3281051076), the `onClick` handlers can be safely replaced with the `handleClick` callback.
```suggestion
        let 
        if (expandable) {
            return {
                role: 'button' as AriaRole,
                tabIndex: 0,
                onClick: handleClick,
                onKeyDown: onKeyDown,
                'aria-controls': cardId,
                'aria-expanded': showContent,
            };
        }
        if (onClick) {
            return {
                role: 'button' as AriaRole,
                tabIndex: 0,
                onClick: handleClick,
                onKeyDown,
            };
        }
        return {};
    }, [expandable, onClick, showContent, cardId, handleClick, onKeyDown]);
```

src/components/external/CapitalOffer/components/TermSelector.tsx
```diff
@@ -0,0 +1,121 @@
...
+    );
+
+    const selectTerm = useCallback(
+        (term: number) => {
+            onTermSelect(term);
+            userEvents.addEvent?.('Selected repayment term', {
+                ...sharedAnalyticsEventProperties,
+                label: 'Term selected',
+                value: term,
+            });
```
glad-adyen, line 45: I don't see any integration test to confirm that this event is sent as expected.

src/components/external/CapitalOffer/components/TermSelector.tsx
```diff
@@ -0,0 +1,121 @@
...
+                            key={term}
+                            noOutline
+                            noPadding
+                            onClick={!isDisabled ? () => selectTerm(term) : undefined}
+                            classNameModifiers={[
+                                'adyen-pe-capital-offer-selection__term',
+                                ...(isSelected ? ['adyen-pe-capital-offer-selection__term--selected'] : []),
+                                ...(isDisabled ? ['adyen-pe-capital-offer-selection__term--disabled'] : []),
+                            ]}
+                        >
```
glad-adyen, line 79: A `"radiogroup"` role on the parent element requires that the child elements each have a `"radio"` role. I'd suggest reviewing the role and attributes relationships here.

src/components/external/CapitalOffer/components/CapitalOfferSummary/CapitalOfferSummary.scss
```diff
@@ -1,14 +1,43 @@
...
+            flex-direction: column;
+            gap: style.token(spacer-060);
+        }
+
+        &-item {
+            display: flex;
+            flex-direction: column;
+            align-items: center;
+            flex: 1;
+        }
```
glad-adyen, line 28: Looks like `stylelint` checks are broken again, it is supposed to report that the CSS properties here are not listed in alphabetical order. You could try running it locally against your affected files to address lint issues.

src/components/internal/Card/types.ts
isuert: Good catch! It's not being used, will remove it.

src/components/internal/Card/Card.tsx
isuert: For the expandable card, CARD_EXPANDABLE_CLASS is being used. It shares some styles with clickable class. We can refactor but I think it's good as it is given that we're moving to Bento soon.

glad-adyen: In that case, I'd suggest this, since it aligns with the rationale that expandable is clickable:
```suggestion
                [CARD_CLICKABLE]: !!onClick || expandable,
```

src/components/internal/Card/Card.tsx
glad-adyen: Indeed. Although the reason for removing it here is because the `if` condition is redundant, hence the callback should no longer depend on `expandable`.

src/components/external/CapitalOffer/components/TermSelector.tsx
isuert: Good catch!

german664 (dismissed):

german664 (approved):

PR #1184, opened 2026-05-25
feat: deprecation warning for availableTranslations option

glad-adyen:
## Summary
This PR introduces a deprecation warning for the `availableTranslations` Core option. This option is currently ignored, and is marked for removal in the next major version.

### DIFF
--- a/packages/shared/core/src/Core.ts
+++ b/packages/shared/core/src/Core.ts
@@ -1,4 +1,4 @@
-import { EMPTY_OBJECT } from '@integration-components/utils';
+import { EMPTY_OBJECT, hasOwnProperty } from '@integration-components/utils';
 import { AuthSession } from './session/AuthSession';
 import Localization from './Localization';
 import { Assets, AssetOptions } from './Assets/Assets';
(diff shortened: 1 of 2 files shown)


packages/shared/core/src/Core.ts
```diff
@@ -91,6 +95,11 @@ export class Core<AvailableTranslations extends TranslationSourceRecord[] = [],
             this.applyAnalyticsOptions();
         }
 
+        if (!this.hasWarnedAboutAvailableTranslationsDeprecation && hasOwnProperty(this.options, 'availableTranslations')) {
```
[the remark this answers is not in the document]
glad-adyen: We're technically never in dev environment for the end-user. So will leave as is for now.

german664 (dismissed):

german664 (approved):

packages/shared/core/src/Core.ts
```diff
@@ -91,6 +95,11 @@ export class Core<AvailableTranslations extends TranslationSourceRecord[] = [],
...
+            console.warn(AVAILABLE_TRANSLATIONS_DEPRECATION_WARNING);
+            this.hasWarnedAboutAvailableTranslationsDeprecation = true;
+        }
+
```
samuelsa-adyen, line 102: Wdyt about adding this to the constructor, and that way we wouldn't have to use the `hasWarnedAboutAvailableTranslationsDeprecation` conditional?

glad-adyen: The constructor is for one-off initialization. If the `availableTranslations` option was not provided from the on set, we would have missed our only opportunity to write this warning.

Imagine that the option was later provided using the `update()` method, with the current approach, we would never miss it, regardless of when and how it was provided.

samuelsa-adyen: Fair enough

samuelsa-adyen (approved):

PR #1194, opened 2026-05-28
Cleanup src folder pt. 1

german664:
Cleanup of stale files in `src`


packages/shared/core/src/preact/types.ts
glad-adyen, line 1: This is a new file. What's the reason for creating it?

glad-adyen (dismissed):

glad-adyen (dismissed):

packages/shared/core/src/preact/types.ts
german664: It was created to host the Preact element and form types after deleting the old root `src/components/types.ts` file. It is a new file and not a moved one because here are only the preact related types, and the other were moved to the right files.

samuelsa-adyen (dismissed):

glad-adyen (dismissed):

samuelsa-adyen (dismissed):

glad-adyen (approved):

samuelsa-adyen (approved):

PR #1195, opened 2026-05-28
Cleanup src folder pt. 2

german664:


vite.config.ts
```diff
@@ -118,7 +118,8 @@ export default defineConfig(({ mode }) => {
                     'components/internal/**/*.{ts,tsx}',
                     'components/utils/*.{ts,tsx}',
                     'hooks/**/*.{ts,tsx}',
-                    'primitives/**/*.{ts,tsx}',
+                    'packages/shared/core/src/session/SessionContext/**/*.{ts,tsx}',
+                    'packages/shared/utils/src/primitives/**/*.{ts,tsx}',
                     'utils/**/*.{ts,tsx}',
                     'core/**/*.{ts,tsx}',
```
glad-adyen, line 124: I think we have to re-visit this list altogether, as some of the entries no longer exist.

For the new entries, I think the paths can be compressed like so:
```ts
'packages/shared/{core,utils}/src/**/*.{ts,tsx}'
```

german664: Should we do it in a future cleanup PR dedicated to this? @glad-adyen

glad-adyen: Sounds fair

glad-adyen (approved):

PR #1198, opened 2026-05-28
Cleanup src folder pt. 3

german664:
This PR moves Assets and translation files out of `src`


vite.config.ts
```diff
@@ -93,6 +93,11 @@ export default defineConfig(({ mode }) => {
         json: {
             stringify: true,
         },
+        resolve: {
+            alias: {
+                '@integration-components/assets': resolve(__dirname, 'packages/shared/assets/src'),
+            },
+        },
```
glad-adyen: Why do we need this? Don't we already have `@integration-components/assets` as a dependency?

packages/shared/core/src/Core.ts
```diff
@@ -6,7 +6,7 @@ import { getCustomTranslationsAnalyticsPayload } from './EventDispatcher/eventDi
 import { SERVER_SIDE_INITIALIZATION_WARNING, shouldWarnAboutServerSideInitialization } from './runtime';
 import { FALLBACK_ENV, getConfigFromCdn, getDatasetFromCdn, resolveEnvironment } from './utils';
 import type { CoreOptions, onErrorHandler, ResolvedEnvironment } from './types';
-import type { TranslationSourceRecord } from '../../../../src/translations';
+import type { TranslationSourceRecord } from '@integration-components/core/translations';
```
glad-adyen: Why are we importing from `@integration-components/core` from within the core package? This should be a relative import path.

packages/shared/core/src/types.ts
```diff
@@ -1,5 +1,5 @@
 import type { SessionRequest } from './ConfigContext.types';
-import type { CustomTranslations as Translations, TranslationSourceRecord } from '../../../../src/translations';
+import type { CustomTranslations as Translations, TranslationSourceRecord } from '@integration-components/core/translations';
```
glad-adyen: Why are we importing from `@integration-components/core` from within the core package? This should be a relative import path.

packages/shared/core/src/session/utils/sessionAwareComponentAvailability/helpers/componentAvailabilityErrors.ts
```diff
@@ -1,5 +1,5 @@
 import { ExternalComponentType } from '@integration-components/types';
-import { TranslationKey } from '../../../../../../../../src/translations';
+import { TranslationKey } from '@integration-components/core/translations';
```
glad-adyen: Why are we importing from `@integration-components/core` from within the core package? This should be a relative import path.

packages/shared/core/src/Localization/utils.ts
```diff
@@ -1,6 +1,6 @@
 import { DEFAULT_TRANSLATIONS, FALLBACK_LOCALE, SUPPORTED_LOCALES } from './constants/localization';
 import { asPlainObject, EMPTY_OBJECT, hasOwnProperty, isFunction } from '@integration-components/utils';
-import type { CustomTranslations, Locale, TranslationOptions, Translations } from '../../../../../src/translations';
+import type { CustomTranslations, Locale, TranslationOptions, Translations } from '@integration-components/core/translations';
```
glad-adyen: Why are we importing from `@integration-components/core` from within the core package? This should be a relative import path.

packages/shared/core/src/Localization/utils.test.ts
```diff
@@ -1,8 +1,8 @@
 import { formatCustomTranslations, formatLocale, getTranslation, interpolateElement, matchLocale, parseLocale, loadTranslations } from './utils';
 import { describe, expect, test } from 'vitest';
 import { SUPPORTED_LOCALES } from './constants/localization';
-import { translations_dev_assets } from '../../../../../src/translations/local';
-import { TranslationKey } from '../../../../../src/translations';
+import { translations_dev_assets } from '@integration-components/core/translations/local';
+import { TranslationKey } from '@integration-components/core/translations';
```
glad-adyen: Why are we importing from `@integration-components/core` from within the core package? These should be relative import paths.

packages/shared/core/src/Localization/Localization.ts
```diff
@@ -25,8 +25,8 @@ import { createWatchlist } from '@integration-components/utils';
 import { ALREADY_RESOLVED_PROMISE, isNull, isNullish, isUndefined, noop, struct } from '@integration-components/utils';
 import { httpGet } from '../Http/http';
 import { SupportedLocales } from './types';
-import { translations_dev_assets } from '../../../../../src/translations/local';
-import localSwapConfig from '../../../../../src/config/translations/swapConfig.json';
+import { translations_dev_assets } from '@integration-components/core/translations/local';
```
glad-adyen: Why are we importing from `@integration-components/core` from within the core package? This should be a relative import path.

packages/shared/core/src/Localization/Localization.ts
```diff
@@ -14,8 +14,8 @@ import type {
     Translations,
     TranslationSource,
     TranslationSourceRecord,
-} from '../../../../../src/translations';
-import { en_US } from '../../../../../src/translations';
+} from '@integration-components/core/translations';
+import { en_US } from '@integration-components/core/translations';
```
glad-adyen: Why are we importing from `@integration-components/core` from within the core package? These should be relative import paths.

vite.config.ts
german664: good catch ✅

packages/shared/core/src/Core.ts
german664: ✅

packages/shared/core/src/types.ts
german664: ✅

packages/shared/core/src/session/utils/sessionAwareComponentAvailability/helpers/componentAvailabilityErrors.ts
german664: ✅

packages/shared/core/src/Localization/utils.ts
german664: ✅

packages/shared/core/src/Localization/utils.test.ts
german664: ✅

packages/shared/core/src/Localization/Localization.ts
german664: ✅

packages/shared/core/src/Localization/Localization.ts
german664: ✅

scripts/upload-assets/package-assets.sh
glad-adyen, line 1: You have omitted translations swap config packaging here. I recommend doing the same as with the other configs (i.e. disputes, pay by link, etc) — copy the swap config to new location while also preserving the old location for CDN

eslint.config.cjs
```diff
@@ -90,10 +90,12 @@ module.exports = [
                     devDependencies: [
                         'stories/**/*',
                         'playwright.config.ts',
+                        'vite.config.ts',
                         'config/**/*.ts',
                         'envs/**/*.ts',
                         'mocks/**/*.ts',
                         'tests/**/*.{ts,js}',
+                        'src/index.ts',
```
glad-adyen: Why was `src/index.ts` added as a dev dependency for ESLint?

scripts/upload-assets/package-assets.sh
german664: ✅

eslint.config.cjs
german664: To suppress a ESLint warning, but it's fine because it will be fixed in the following PR. So I removed it. ✅

glad-adyen (approved):

PR #1200, opened 2026-05-28
Cleanup src folder pt. 4

german664:
Remove `src` entirely


tsconfig.json
```diff
@@ -10,6 +10,6 @@
         "declarationDir": "./dist/types",
         "types": ["node", "vitest/globals", "@testing-library/jest-dom"]
     },
-    "include": ["src/**/*.ts", "src/**/*.tsx", "src/**/*.d.ts", "tests/**/*.ts"],
+    "include": ["src/**/*.ts", "src/**/*.tsx", "declarations/**/*.d.ts", "tests/**/*.ts"],
```
glad-adyen: I think `"src/**/*.tsx"` can be safely removed now.

src/core/Localization/constants/currency-decimals.ts
glad-adyen, line 1: Can you confirm that we have this file in the shared Core package?

src/core/Localization/constants/currency-codes.ts
glad-adyen, line 1: Can you confirm that we have this file in the shared Core package?

src/core/ConfigContext/session/__testing__/MockSessionTestUtils.ts
glad-adyen, line 1: Can you confirm that we have this file in the shared Core package?

src/core/Errors/ErrorPanel.tsx
glad-adyen, line 1: Can you confirm that we did not move this file to the shared Core package, because I doubt that we ever use it?

src/core/Errors/ErrorPanel.scss
glad-adyen, line 1: Can you confirm that we did not move this file to the shared Core package, because I doubt that we ever use it?

tsconfig.json
german664: ✅

glad-adyen (dismissed):

glad-adyen (approved):

PR #1202, opened 2026-05-29
Add shared Vue composables, CustomDataCell, and core improvements

samuelsa-adyen:
- Add useAccountBalances, useContainerQuery, useDownload, useResponsiveContainer, useUniqueId composables
- Add CustomDataCell shared component for custom data grid cells
- Add formatAmountWithCurrencyCode utility
- Update ConfigProvider to use initialized latch pattern
- Update Storybook Vue container for locale/font reactivity
- Add bento-vue3, ui-assets-icons catalog entries

## Tested scenarios

**Fixed issue**:


packages/shared/composables-vue/src/useUniqueId.ts
```diff
@@ -0,0 +1,11 @@
+import { uniqueId } from '@integration-components/utils';
+
+/**
+ * Vue equivalent of the Preact `useUniqueId` hook.
+ * Returns a stable unique numeric string (no prefix) for use in DOM IDs.
+ */
+export function useUniqueId(): string {
+    return uniqueId().replace(/.*?(?=\d+$)/, '');
+}
```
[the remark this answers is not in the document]
samuelsa-adyen: We will keep the same format as the previous implementation

packages/shared/composables-vue/src/useUniqueId.ts
```diff
@@ -0,0 +1,12 @@
+import { uniqueId } from '@integration-components/utils';
+
+/**
+ * Vue equivalent of the Preact `useUniqueId` hook.
+ * Returns a stable unique numeric string (no prefix) for use in DOM IDs.
+ */
+export function useUniqueId(): string {
+    const id = uniqueId();
+    return id.slice(id.lastIndexOf('-') + 1);
```
glad-adyen: Not deviating much from the existing implementation, can you try this? The rationale is to break the match overlap between `.*` and `\d+`.
```suggestion
    return uniqueId().replace(/^(?:.*\D)?(?=\d+$)/, '');
```

samuelsa-adyen: Added!

german664 (approved):

packages/shared/core/src/Localization/amount/amount-util.ts
```diff
@@ -72,6 +72,11 @@ export const formatAmountWithoutCurrency = (locale: string, options: Intl.Number
         .trim();
 };
 
+export const formatAmountWithCurrencyCode = (amount: number, locale: string, currencyCode: string): string => {
+    const localisedAmount = getLocalisedAmount(Math.abs(amount), locale, currencyCode, true, { currencyDisplay: 'symbol' });
+    return `${amount < 0 ? `- ${localisedAmount}` : localisedAmount} ${currencyCode}`;
```
glad-adyen, line 77: This looks like the amount will be rendered with both currency symbol (from `localisedAmount`) and currency code (in the template literal) — something like **$125.99 USD**. Is this intended, or is it not the case here?

packages/shared/composables-vue/src/CustomDataCell.ts
```diff
@@ -0,0 +1,51 @@
...
+                return h('span', { class: data.config?.className }, String(data.value ?? ''));
+            }
+            if (data.type === 'button' && data.config) {
+                return h(
+                    BentoButton,
+                    {
+                        variant: 'secondary',
+                        class: data.config.className,
+                        onClick: (e: Event) => {
+                            e.stopPropagation();
```
glad-adyen, line 38: Just curious. Why stop the propagation of the event here?

samuelsa-adyen: The button is rendered inside a DataGrid row cell. Without stopPropagation, clicking the button would also trigger the row's click handler.

packages/shared/core/src/Localization/amount/amount-util.ts
samuelsa-adyen: I tested this and the amount is actually rendered correctly due to the `hideCurrency` parameter being sent to the `getLocalisedAmount` function. We are in the end following this format:

image

glad-adyen (approved):

PR #1203, opened 2026-05-29
Feature/variable term improvements

isuert:
- Lifts selectedAmount and selectedTerm state from CapitalOfferSelection into the parent CapitalOffer component, so the user's slider position and repayment term are preserved when navigating back from the review screen
- The default term selection logic was updated to prefer 180 days (6 months) when available


src/components/external/CapitalOffer/components/CapitalOfferSelection/CapitalOfferSelection.tsx
german664: Should be include a check to see if `matchedOffer`  is available?

I think If the user clicks this button when `matchedOffer` is `undefined` the onReview function will trigger, but it will skip the mutation.

glad-adyen (approved):

german664 (approved):

PR #1207, opened 2026-06-01
feat(payouts,reports): refactor to use shared CustomDataCell and catalog deps

samuelsa-adyen:
## Tested scenarios

**Fixed issue**:


packages/domains/payouts/vue/src/PayoutsOverview/components/PayoutsTable.vue
```diff
@@ -94,7 +94,7 @@ const columns = computed<BentoColumn[]>(() => {
             const labelKey = `payouts.overview.list.fields.${key}`;
             cols.push({
                 field: key,
-                label: i18n.has(labelKey) ? i18n.get(labelKey) : key,
+                label: i18n.has(labelKey) ? i18n.get(labelKey) : i18n.get(key as any),
```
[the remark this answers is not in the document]
samuelsa-adyen: This defaults to the key which is the intended behavior.

packages/domains/payouts/vue/src/PayoutsOverview/styles/PayoutsTable.scss
```diff
@@ -2,6 +2,7 @@
 
 .adyen-pe-payouts-table {
     position: relative;
+    container-type: inline-size;
```
glad-adyen: Not sure why stylelint didn't catch this, but I'm sure there is a rule to enforce that `container-type` style declaration comes before `position`.

packages/domains/payouts/vue/src/PayoutDetails/components/PayoutData.vue
```diff
@@ -128,9 +129,8 @@ function formatPayoutDate(dateStr: string): string {
     return dateFormat(dateStr, DATE_FORMAT_PAYOUT_DETAILS);
 }
 
-function formatAmount(amount: { value: number; currency: string }): string {
-    return `${i18n.amount(amount.value, amount.currency, { hideCurrency: true })} ${amount.currency}`;
-}
+const formatAmount = (amount: { value: number; currency: string } | null | undefined) =>
+    amount ? formatAmountWithCurrencyCode(amount.value, i18n.locale, amount.currency) : '';
```
glad-adyen: What do you think about expecting a non-nullable amount here. And using `v-if` in the template to render the formatted amount conditionally, just like you are doing for created date currently.

packages/domains/reports/vue/src/ReportsOverview/components/ReportsTable.vue
```diff
@@ -203,6 +202,15 @@ function handleItemsPage(size: number) {
     props.updateLimit?.(size);
 }
 
+function formatDate(dateStr: string): string {
+    return i18n.date(dateStr, {
+        year: 'numeric',
+        month: 'short',
+        day: 'numeric',
+        timeZone: 'UTC',
+    });
```
glad-adyen: Isn't this format reused in several places? Would be nice to share this definition.

packages/domains/payouts/vue/src/PayoutsOverview/components/PayoutsTable.vue
glad-adyen: Let's keep an eye on this pattern here, because of dynamic keys and fallback behavior. You can compare with how these are handled for the Preact components.

packages/domains/payouts/vue/src/PayoutsOverview/styles/PayoutsTable.scss
samuelsa-adyen: Changed ✅

packages/domains/payouts/vue/src/PayoutDetails/components/PayoutData.vue
samuelsa-adyen: Changed ✅

packages/domains/reports/vue/src/ReportsOverview/components/ReportsTable.vue
samuelsa-adyen: Changed for reusable `DATE_FORMAT_REPORTS` ✅

packages/domains/payouts/vue/src/PayoutsOverview/components/PayoutsTable.vue
samuelsa-adyen: Sure, this fix actually makes sure the custom strings that are passed by the platforms are picked up.

Fixed the type as well, made it `TranslationKey`

glad-adyen (approved):

german664 (approved):

PR #1224, opened 2026-06-02
Resolve issues with release workflow

glad-adyen:
## Summary
This PR introduces changes to resolve these remaining issues with the release workflow:

- GitHub release remains as draft, until manually published
- SRI hashes workflow fails when triggered by the `tag-and-release` workflow

### DIFF
--- a/.github/workflows/tag-and-release.yml
+++ b/.github/workflows/tag-and-release.yml
@@ -75,8 +75,8 @@ jobs:
           git tag -d "$SCOPED_TAG"
           git push origin --delete "$SCOPED_TAG"
 
-          # Update the GitHub release to point to the new tag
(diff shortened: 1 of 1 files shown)


german664 (dismissed):

.github/workflows/tag-and-release.yml
```diff
@@ -129,7 +129,7 @@ jobs:
   generate-sri-hashes:
```
samuelsa-adyen: Should we just remove this step? Seems like at the time it is not necessary given that we still need to upload the assets manually before this.

glad-adyen: You're right — I would comment it for now.

glad-adyen: Done ✅

samuelsa-adyen (approved):

german664 (approved):

PR #1235, opened 2026-06-11
Feature/introduce capital state

isuert:
Replaces the dual-endpoint approach (`getDynamicGrantOffersConfiguration` + `getGrants`) with a single `getCapitalState` endpoint to determine capital component state.

- Updated CapitalStateResourceV1.ts to the new endpoint shape (hasGrants, renewableGrants, dynamicOffer)
- Extracted state resolution into a new getCapitalState utility and exposed getState on CapitalOffer component
- CapitalOverview now fetches capital state first, then conditionally fetches grants only when hasGrants is true
- Updated mocks, MSW handlers, and renamed integration test specs to match new scenarios


packages/domains/capital/preact/src/utils/capital/getCapitalState.ts
```diff
@@ -0,0 +1,35 @@
...
+        return {
+            hasGrants: false,
+            hasOffer: false,
+            hasRenewableGrants: false,
+            state: 'isInUnsupportedRegion',
+        };
+    }
+
+    const { getCapitalState } = session.context.endpoints;
+    const capitalStateResponse = await getCapitalState?.(EMPTY_OBJECT, { query: EMPTY_OBJECT }).catch(() => undefined);
```
german664, line 21: Why are we returning `undefined` here in the catch block? In that case every API failure becomes undefined, meaning that we could have silent failures and no way of handling it wherever this is being consumed

packages/domains/capital/preact/tests/integration/capitalOffer/unqualified.spec.ts
```diff
@@ -16,10 +16,3 @@ test.describe('Error - Dynamic offer config - No config', () => {
         await expect(page.getByRole('button', { name: 'Reach out to support' })).toBeHidden();
     });
 });
-
```
german664, line 19: Why are we removing this test?

packages/domains/capital/preact/tests/integration/capitalOffer/unqualified.spec.ts
```diff
@@ -16,10 +16,3 @@ test.describe('Error - Dynamic offer config - No config', () => {
...
-test.describe('onContactSupport argument', () => {
```
german664, line 20: Why are we removing this test?

packages/shared/types/src/api/resources/CapitalStateResourceV1.ts
```diff
@@ -4,7 +4,7 @@
  */
 
 export interface paths {
-    '/v1/capital/capitalStates': {
+    '/v1/capital/capitalState': {
```
glad-adyen, line 7: When was this endpoint introduced? If older versions of the SDK rely on this endpoint, then I worry that this change its name isn't backwards compatible.

packages/shared/types/src/api/resources/CapitalStateResourceV1.ts
```diff
@@ -37,12 +37,10 @@ export interface components {
...
             value: number;
         };
-        /** @enum {string} */
-        CapitalState: 'NOT_ELIGIBLE' | 'OFFER_ELIGIBLE' | 'REQUESTED_OR_ACTIVE_GRANT' | 'EARLY_RENEWAL_ELIGIBLE';
         CapitalStateDTO: {
-            activeOrRequestedGrants: components['schemas']['GrantResponseDTO'][];
-            capitalState: components['schemas']['CapitalState'];
             dynamicOffer?: components['schemas']['DynamicOffersResponseV2DTO'];
+            hasGrants: boolean;
+            renewableGrants: components['schemas']['GrantResponseDTO'][];
```
glad-adyen, line 43: Same backwards compatibility concerns with this change to the response shape.

packages/domains/capital/preact/src/utils/capital/getCapitalState.ts
glad-adyen: The error is suppressed here, because the goal is state detection, rather than propagating error. It matches the semantics of the previous behavior.

packages/domains/capital/preact/src/CapitalOffer/components/CapitalOffer/CapitalOffer.tsx
```diff
@@ -22,37 +22,27 @@ const sharedAnalyticsEventProperties = {
...
+        fetchOptions: { enabled: !externalCapitalState && !!getCapitalState },
         queryFn: useCallback(async () => {
-            return getDynamicGrantOffersConfiguration?.(EMPTY_OBJECT, { query: EMPTY_OBJECT });
-        }, [getDynamicGrantOffersConfiguration]),
+            return getCapitalState?.(EMPTY_OBJECT, { query: EMPTY_OBJECT });
+        }, [getCapitalState]),
     });
 
-    const config = externalDynamicOffersConfig || internalDynamicOffersConfig;
+    const state = externalCapitalState || internalCapitalState;
```
samuelsa-adyen, line 46: I think all of this logic is to get the `state` and `capitalStateError` vars, if we aren't using them in this component but moving them down, is there any issue with just fetching this in the `CapitalOfferSelection` component?

packages/domains/capital/preact/src/CapitalOffer/components/CapitalOfferSelection/CapitalOfferSelection.tsx
```diff
@@ -332,14 +331,16 @@ export const CapitalOfferSelection = ({
 
     const hasSingleTerm = useMemo(() => allTerms.length === 1, [allTerms.length]);
 
+    const isUnqualified = useMemo(() => capitalState !== undefined && !capitalState.dynamicOffer, [capitalState]);
+
```
samuelsa-adyen, line 335: This component is huge, could we consider moving over some of the logic to a hook/external function?

packages/shared/types/src/api/resources/CapitalStateResourceV1.ts
isuert: This endpoint is not released yet - no need to worry about backwards compatibility.

packages/shared/types/src/api/resources/CapitalStateResourceV1.ts
isuert: Same answer as above

packages/domains/capital/preact/src/CapitalOffer/components/CapitalOffer/CapitalOffer.tsx
isuert: But we are using them here as well.

packages/domains/capital/preact/src/CapitalOffer/components/CapitalOfferSelection/CapitalOfferSelection.tsx
isuert: I agree but I think that's out of the scope of this PR.

packages/domains/capital/preact/tests/integration/capitalOffer/unqualified.spec.ts
isuert: Reverted.

packages/shared/types/src/api/resources/CapitalStateResourceV1.ts
glad-adyen: As for being released, this endpoint was released in `v1.12.1`, except you are saying it has never been used until now. So are you saying it has never been used until now?

german664 (dismissed):

packages/shared/types/src/api/resources/CapitalStateResourceV1.ts
isuert: Indeed, that's what I mean - even though the endpoint is in the resources file, there is no code in the current releases that uses this endpoint.

glad-adyen (approved):

PR #1259, opened 2026-06-16
Fix release workflows

glad-adyen:
## Summary
This PR includes the following changes and updates to fix the current release workflows.

- Exclude internal (`@integration-components`) packages from stable release
- Open PR from intermediate branch to update `develop` with `main` after release
- Skip drafts for GitHub releases and release immediately

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -97,6 +97,12 @@ jobs:
               git checkout -b "$PRERELEASE_BRANCH"
             fi
 
+            # Strip internal workspace packages from pre.json so that only the
+            # root package is bumped when exiting pre-release mode.
+            jq '.initialVersions |= with_entries(select(.key == "@adyen/adyen-platform-experience-web"))' \
+              .changeset/pre.json > .changeset/pre.json.tmp \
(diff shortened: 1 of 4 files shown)


german664 (dismissed):

.github/workflows/prepare-release.yml
```diff
@@ -71,30 +71,33 @@ jobs:
           # Enter pre-release mode if not already active
           if [ ! -f .changeset/pre.json ]; then
             pnpm changeset pre enter "$TAG"
+            jq '.initialVersions |= with_entries(select(.key == "@adyen/adyen-platform-experience-web"))' \
+              .changeset/pre.json > .changeset/pre.json.tmp \
+              && mv .changeset/pre.json.tmp .changeset/pre.json
```
glad-adyen: I'd say we skip this rewrite here, in favor of the one during the promotion step, since that is the point when it really matters. That will also ensure that we don't interfere with changesets behaviors or expectations for pre-releases.

german664: Good point! ✅

german664 (approved):

samuelsa-adyen (approved):

PR #1266, opened 2026-06-18
chore: Add description to filters & fix container queries

samuelsa-adyen:
## Tested scenarios

**Fixed issue**:

### DIFF
--- a/packages/domains/payouts/vue/src/PayoutsOverview/components/PayoutsFilters.vue
+++ b/packages/domains/payouts/vue/src/PayoutsOverview/components/PayoutsFilters.vue
@@ -118,6 +118,7 @@ const filterConfig = computed<BentoFilterBarModel>(() => {
                 listboxItems: props.balanceAccounts.map((a: IBalanceAccountBase) => ({
                     label: a.description || a.id,
                     value: a.id,
+                    description: a.description ? a.id : undefined,
                 })),
             },
         });
--- a/packages/shared/composables-vue/src/useContainerQuery.ts
+++ b/packages/shared/composables-vue/src/useContainerQuery.ts
@@ -1,8 +1,10 @@
-import { ref, computed, watch, onUnmounted, type Ref, type ComputedRef } from 'vue';
+import { ref, computed, watch, onUnmounted, inject } from 'vue';
+import { COMPONENT_REF_KEY } from '@integration-components/core/vue/Context/constants';
 
 type ContainerQuery = readonly [string, number, { min?: number; max?: number }?];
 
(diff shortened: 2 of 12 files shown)


german664 (approved):

packages/shared/composables-vue/src/useContainerQuery.ts
```diff
@@ -1,8 +1,10 @@
-import { ref, computed, watch, onUnmounted, type Ref, type ComputedRef } from 'vue';
+import { ref, computed, watch, onUnmounted, inject } from 'vue';
+import { COMPONENT_REF_KEY } from '@integration-components/core/vue/Context/constants';
 
 type ContainerQuery = readonly [string, number, { min?: number; max?: number }?];
 
-export function useContainerQuery<T extends ContainerQuery>(containerRef: Ref<HTMLElement | null>, query: T): ComputedRef<boolean> {
+export const useContainerQuery = <T extends ContainerQuery>(query: T) => {
+    const containerRef = inject(COMPONENT_REF_KEY, ref<HTMLElement | null>(null));
```
glad-adyen, line 7: By injecting `COMPONENT_REF_KEY` here as `containerRef`, you limit the `useContainerQuery` composable to the root container of external components, making it difficult to use it for responsive containers anywhere else down the render tree.

samuelsa-adyen: That's the point, no? In Preact we're doing 
`const { componentRef } = useCoreContext();` 
Which gets the componentRef for the root container of the external components as well.

packages/domains/payouts/vue/src/PayoutsOverview/components/PayoutsFilters.vue
```diff
@@ -118,6 +118,7 @@ const filterConfig = computed<BentoFilterBarModel>(() => {
                 listboxItems: props.balanceAccounts.map((a: IBalanceAccountBase) => ({
                     label: a.description || a.id,
                     value: a.id,
+                    description: a.description ? a.id : undefined,
```
glad-adyen, line 121: For a balance account without description (only id), how will Bento render the listbox option? Just want to be sure it will not render the balance account id twice.

samuelsa-adyen: image

Looks like this, single mention of the ID

packages/shared/composables-vue/src/useContainerQuery.ts
glad-adyen: Interesting — this limitation also applies to Preact. Will keep this in mind.

glad-adyen (approved):

PR #1267, opened 2026-06-19
Vue integration tests for payouts domain 

glad-adyen:


packages/domains/payouts/vue/tests/integration/PayoutsOverview/default.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+
+            const nextDateRangePreset = 'Year to date';
+
+            // (1) Date preset select
+            await expect(dateRangePresetSelectDialog).toBeHidden();
+            await expect(dateRangePresetSelectButton).toBeVisible();
+            await expect(dateRangePresetSelectButton).toHaveText('Last 30 days');
+
+            // (2) Calendar
+            // await expect(todayDate).toBeVisible();
```
german664, line 127: Why are this tests commented out?

glad-adyen: A couple reasons:

- The Bento datepicker component does not have a definite annotation for today date that can be used to define an equivalent locator (unlike in the Preact datepicker where we can use `[data-today='1'][aria-selected='true']`).
- The Bento datepicker does not always render the today date initially (it renders from the start of the selected range).

german664: So should we remove them instead?

glad-adyen: I am preserving them for now, because I am considering extracting these tests across all components much later to avoid excessive repetitions per domain. For that, I need to know how the Vue tests compare to the Preact variants to know what to abstract.

Sounds good?

german664: Sounds great ✅

german664 (dismissed):

german664 (approved):

PR #1268, opened 2026-06-26
Feature/introduce early renewals

isuert:
## Summary
Adds support for the early renewal flow in the Capital Offer and Capital Overview components.


packages/domains/capital/preact/src/CapitalOffer/components/CapitalOffer/CapitalOffer.tsx
```diff
@@ -42,7 +43,7 @@ const DynamicCapitalOffer: FunctionalComponent<ExternalUIComponentProps<CapitalO
         }, [getCapitalState]),
     });
 
-    const state = externalCapitalState || internalCapitalState;
+    const state = useMemo(() => externalCapitalState || getEnhancedCapitalState(internalCapitalState), [externalCapitalState, internalCapitalState]);
```
german664: `getEnhancedCapitalState(undefined)` always returns a not `undefined` object `{ hasGrants: false, dynamicOffer: undefined, renewableGrants: [] }` and `CapitalOfferSelection` decides if the user is "unqualified" from that object, so  the initial render now evaluates `isUnqualified === true` and renders with `emptyGrantOffer` until the fetch resolves and the slider appears.

I would suggest to do something like 

```
const state = useMemo(
    () => externalCapitalState ?? (internalCapitalState && getEnhancedCapitalState(internalCapitalState)) ?? undefined,
    [externalCapitalState, internalCapitalState]
);
```

So we don't have that wrong intermediate state while it fetches.

isuert: Good catch! Noticed the same - it should be fixed now.

german664 (dismissed):

packages/shared/ui-components-preact/src/Slider/Slider.tsx
```diff
@@ -4,8 +4,13 @@ import { JSX } from 'preact';
 import { calculateProgress } from './calculateProgress';
 import cx from 'classnames';
 
+export function calculateSliderAdjustedMinValue(values: number[], step: number) {
```
glad-adyen: The name of this function `calculateSliderAdjustedMinValue` is a bit misleading. It clearly computes an adjusted max value from the list of numbers. However, I see from how it is used that it is doing this computation for the minimum grant amount (using the minimum repayment amount, etc), hence the misleading name.

I'd suggest that the name of the function itself better reflect its behavior and output to ensure there is no ambiguity. It can simply be renamed to `calculateAdjustedMaxValue` irrespective of the context of its usage.

isuert: But then wouldn't it be misleading in the consumer end, i.e. calling a function that gives max value, for the min value of the slider? I actually like that in the current implementation there is consistency between calculateSliderAdjustedMinValue and calculateSliderAdjustedMidValue.

glad-adyen: On the contrary, the consumer shouldn't inform the name of the function, but rather what the function actually does. For the consumer, you have variables for that (a more descriptive variable name should suffice).

As an analogy, imagine that the function `Math.min()` returns the maximum value from a list of numbers, that will be very misleading.

My suggestion is:
- For the function, rename to something like `calculateAdjustedMaxValue`, since that aligns with the expected result from the function.

- For the usage, you can use a more descriptive variable name to describe the computation, e.g. `capitalSliderMinValue`.

glad-adyen (dismissed):

glad-adyen (dismissed):

glad-adyen (approved):

PR #1290, opened 2026-07-13
feat(payByLink): add shared Vue migration foundation

german664:
## Summary

- add framework-neutral domain modules for Payment Link Creation, Details, and Overview
- rewire the existing Preact components to consume the shared contracts, constants, and utilities
- centralize the cross-component `StoreIds` type
- add the Pay By Link Vue package entrypoint and shared Vue dependencies
- type Vue element analytics names as `ExternalComponentType` and align migrated domains with their SDK identifiers
- preserve details timestamp formatting and validate creation payload and fixed-validity behavior with unit tests


packages/shared/assets/src/translations/en-US.json
```diff
@@ -150,6 +150,7 @@
     "common.errors.errorCode": "The error code is %{requestId}",
     "common.errors.errorCodeSupport": "Contact support for help and share error code %{requestId}",
     "common.errors.fieldRequired": "This field is required",
+    "common.errors.maxLength": "Must be at most %{maxLength} characters long.",
```
glad-adyen, line 153: This seems new. Did we have it before? Might be kinda weird having to request new translations as a result of migrations.

german664: Discussed offline ✅

glad-adyen (approved):

PR #1304, opened 2026-07-22
Refactor capital region retrieval

isuert:
Moves capital region retrieval from setup endpoint to capital state endpoint. Adds stories and tests for CA and US.


packages/domains/capital/domain/src/config/supportedRegions.json
```diff
@@ -0,0 +1 @@
+["AU", "CA", "EU", "GB", "US"]
```
glad-adyen, line 1: Why do we need this as a JSON file, rather than just exporting an array constant?

glad-adyen (dismissed):

packages/domains/capital/domain/src/config/supportedRegions.json
isuert: Yes, if we want to expand in a new region and only the subtitle changes we can update this file and the translations without having a new release.

glad-adyen: But that will still require a new release, since I don't see anywhere this file is being pulled from CDN. To achieve the desired goal, you have to also pull this config from CDN with a local fallback, like we do for other configs.

isuert: Good catch! Addressed.

glad-adyen (dismissed):

packages/domains/capital/preact/src/internal/CapitalHeader/helpers.ts
```diff
@@ -1,47 +1,18 @@
...
+        case 'GB':
             return 'capital.common.loanProviderInfo.GB' satisfies TranslationKey;
-        case SupportedLocation.US:
+        case 'US':
             return 'capital.common.loanProviderInfo.US' satisfies TranslationKey;
-        case SupportedLocation.CA:
+        case 'CA':
             return 'capital.common.loanProviderInfo.CA' satisfies TranslationKey;
         default:
             return null;
```
ElifDogaliBal: As a follow up question on the Glad's comment, in the case of the supported regions expanded and added to supportedRegions.json‎ and the subtitle text added to translations, the subtitle will not be rendered for new region since the default is null here. To support the new regions the default can be changed as `capital.common.loanProviderInfo.${region}`

isuert: You're right! Addressed.

glad-adyen (approved):

ElifDogaliBal (approved):

PR #1308, opened 2026-07-30
Address inconsistent error message display

glad-adyen:


packages/domains/payByLink/domain/src/index.ts
```diff
@@ -1,7 +1,7 @@
...
 export * from './constants';
 export * from './types';
-export * from './PaymentLinkCreation';
-export * from './PaymentLinkDetails';
-export * from './PaymentLinksOverview';
-export * from './PaymentLinkSettings';
+export * from './PaymentLinkCreation/index';
+export * from './PaymentLinkDetails/index';
+export * from './PaymentLinksOverview/index';
+export * from './PaymentLinkSettings/index';
```
samuelsa-adyen, line 7: Why do we need to be explicit about this?

packages/shared/composables-vue/src/getErrorMessage.ts
```diff
@@ -0,0 +1,68 @@
...
+    messages: ['common.errors.contactSupport'],
+};
+
+const getCommonErrorMessage = (
+    errorCode: string | undefined,
+    notFoundMessage: TranslationKey,
+    onContactSupport?: () => void
+): ErrorMessageInfo | null => {
+    switch (errorCode) {
+        case '29_001':
```
samuelsa-adyen, line 25: We are repeating these constants a bunch of times, could it be a good idea to put them in an enum or constants object?

samuelsa-adyen (approved):

packages/domains/payByLink/domain/src/index.ts
glad-adyen: The `vue-docgen-api` package (used by Vue Storybook) is unable to resolve the index module if not specified directly. Without these direct index paths, it shows several warnings when the storybook server is started.

packages/shared/composables-vue/src/getErrorMessage.ts
glad-adyen: Which of the constants are you referring to? The error codes or translation keys?

isuert (approved):

PR #1309, opened 2026-07-30
Address inconsistent component lifecycle

glad-adyen:
## Summary
This PR introduces changes for improved and consistent component and session lifecycle across Preact and Vue. It also addresses some inherent reactivity issues for better performance.

### DIFF
--- a/packages/shared/core/src/vue/UIElement.ts
+++ b/packages/shared/core/src/vue/UIElement.ts
@@ -1,8 +1,14 @@
-import { createApp, h, reactive, type App, type Component } from 'vue';
+import { createApp, h, reactive, ref, type App, type Component } from 'vue';
 import { createI18n as createVueI18n } from 'vue-i18n';
 import type { ExternalComponentType } from '@integration-components/types';
 import { uuid } from '@integration-components/utils';
 import UIElementProvider from './UIElementProvider.vue';
+
+export const createRefreshContext = () => {
+    const refreshCount = ref(0);
+    const refresh = () => void refreshCount.value++;
+    return { refresh, refreshCount };
+};
 
 /**
  * Base class that mirrors the Preact BaseElement/UIElement mount/update/unmount lifecycle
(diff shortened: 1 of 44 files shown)


packages/shared/core/src/vue/UIElement.ts
```diff
@@ -1,9 +1,15 @@
-import { createApp, h, reactive, type App, type Component } from 'vue';
+import { createApp, h, reactive, ref, type App, type Component } from 'vue';
 import { createI18n as createVueI18n } from 'vue-i18n';
 import type { ExternalComponentType } from '@integration-components/types';
 import { uuid } from '@integration-components/utils';
 import UIElementProvider from './UIElementProvider.vue';
 
+export const createRefreshContext = () => {
+    const refreshCount = ref(0);
+    const refresh = () => void refreshCount.value++;
```
samuelsa-adyen, line 9: Could we remove `void` from here?

samuelsa-adyen:
This seems a big change for components + interconnectivity between them, could we add a changeset to this?

glad-adyen:
> This seems a big change for components + interconnectivity between them, could we add a changeset to this?

I don't exactly have a strong opinion about this one, because under the hood it just translates to a component refresh optimization for the existing Preact components, while keeping the same consistency for their Vue counterparts.

However, if we feel strongly about this, then we can add a change entry before next release.

packages/shared/core/src/vue/UIElement.ts
glad-adyen: We can indeed. I already know this makes SonarQube unhappy.

ElifDogaliBal (approved):

packages/shared/core/src/vue/UIElement.ts
glad-adyen: I have taken note of this, and it will be changed in a follow-up [PR](https://github.com/Adyen/adyen-platform-experience-web/pull/1321). I do not want to block the current PR stack.

samuelsa-adyen (approved):

PR #1313, opened 2026-08-03
chore(payouts): migrate Vue styles to CSS Modules

samuelsa-adyen:
## Tested scenarios

**Fixed issue**:


isuert (approved):

packages/domains/payouts/vue/src/PayoutDetails/components/PayoutDetailsContainer.vue
```diff
@@ -75,7 +75,7 @@ const showLoadingPlaceholder = computed(() => isFetching.value && !data.value &&
 
 <template>
     <div class="adyen-pe-payout-details">
```
glad-adyen: Why are we retaining this classname?

packages/domains/payouts/vue/src/PayoutDetails/components/PayoutDetailsContainer.module.scss
```diff
@@ -0,0 +1,12 @@
...
+
+.loading {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 100%;
+}
+
+.errorContainer {
+    @include mixins.adyen-pe-error-message-display-container;
```
glad-adyen: Does this mixin include only CSS declarations?

packages/domains/payouts/vue/src/PayoutDetails/components/PayoutData.module.scss
```diff
@@ -0,0 +1,92 @@
+@use '@integration-components/style' as style;
+
+.root {
+    --details-padding: #{style.token(spacer-070)};
```
glad-adyen: I remember there are some styles (contexts) that override this custom property. Would be good to check them out to be sure the expected behavior remains obtainable.

packages/domains/payouts/vue/src/PayoutDetails/components/PayoutDetailsContainer.module.scss
samuelsa-adyen: It does not - I just made the selector dynamic, so it defaults to the current class but we'll send the global specifier for the new modules. Once everything is migrated, we could change the mixin

packages/domains/payouts/vue/src/PayoutDetails/components/PayoutData.module.scss
samuelsa-adyen: Good catch - Returned back to `--adyen-pe-data-details-padding` for consistency

packages/domains/payouts/vue/src/PayoutDetails/components/PayoutDetailsContainer.vue
samuelsa-adyen: Removed.

glad-adyen (approved):

german664 (approved):

PR #1321, opened 2026-08-05
Address inconsistent component filters and pagination results

glad-adyen:

### DIFF
--- a/packages/domains/transactions/preact/src/TransactionsOverview/hooks/useTransactionsList.ts
+++ b/packages/domains/transactions/preact/src/TransactionsOverview/hooks/useTransactionsList.ts
@@ -1,9 +1,9 @@
-import { isFunction, hasCustomField, mergeRecords, normalizeCustomFields } from '@integration-components/utils';
+import { isFunction, hasCustomField, listFrom, mergeRecords, normalizeCustomFields } from '@integration-components/utils';
 import { ITransaction, CustomDataRetrieved } from '@integration-components/types';
 import { TransactionsFilters } from '../types';
 import { DEFAULT_PAGE_LIMIT, LIMIT_OPTIONS } from '@integration-components/ui-components-preact/Pagination/constants';
 import { TRANSACTION_FIELDS, TRANSACTION_FIELDS_REMAPS } from '../../../../domain/src';
-import { getTransactionsFilterParams, getTransactionsFilterQueryParams } from '../components/utils';
+import { getTransactionsFilterParams } from '../components/utils';
 import type { TransactionsOverviewComponentProps } from '../types';
 import { useCursorPaginatedRecords } from '@integration-components/ui-components-preact/Pagination/hooks';
 import { useCustomColumnsData } from '@integration-components/hooks-preact';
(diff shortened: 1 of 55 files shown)


packages/domains/transactions/vue/src/TransactionsOverview/components/TransactionFilters/TransactionsFilters.vue
```diff
@@ -207,36 +220,136 @@ const FILTER_LABELS: Partial<Record<string, FilterType>> = {
...
+});
+
+onMounted(() => {
+    document.addEventListener('click', onFilterBarClick, true);
+});
+
+onUnmounted(() => {
+    transactionsFilterBar.value?.$el?.removeEventListener('click', onFilterBarClick, true);
+    document.removeEventListener('click', onFilterBarClick, true);
+});
```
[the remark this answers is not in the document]
glad-adyen: Approached this entirely differently in the end.

packages/domains/transactions/preact/src/TransactionsOverview/hooks/useTransactionsList.ts
```diff
@@ -29,21 +29,35 @@ const useTransactionsList = ({
...
+
     const fetchTransactions = useCallback(
-        async (requestParams: Pick<Parameters<NonNullable<typeof getTransactions>>[1]['query'], 'limit' | 'cursor'>, signal?: AbortSignal) => {
-            const query: Parameters<NonNullable<typeof getTransactions>>[1]['query'] = {
+        async (paginationParams: typeof filterParams & { cursor?: string; limit: number }, signal?: AbortSignal) => {
+            type TransactionsQuery = Parameters<NonNullable<typeof getTransactions>>[1]['query'];
+            const activeFilterParams = activeFilterParamsRef.current;
+            const hasCurrentFilterParams = (Object.keys(activeFilterParams) as (keyof typeof activeFilterParams)[]).every(
+                key => paginationParams[key] === activeFilterParams[key]
+            );
```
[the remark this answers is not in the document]
glad-adyen: The filter params don't have array values, so this does not apply.

packages/shared/composables-vue/src/useBalanceAccountFilterState.ts
```diff
@@ -0,0 +1,44 @@
...
+interface UseBalanceAccountFilterStateOptions<T extends BalanceAccount> {
+    balanceAccounts: () => readonly T[] | undefined;
+    initialValue?: string;
+    preserveValue?: (value: string | undefined) => boolean;
+}
+
+export function useBalanceAccountFilterState<T extends BalanceAccount>({
+    balanceAccounts,
+    initialValue,
+    preserveValue,
```
ElifDogaliBal, line 17: What do you think getting this preserveValue as an array of objects and mapping all the options for balance account filter together with preserved values so, the balanceAccountOptions object on line 22 can return all of the options? As an example if it returns all the balance accounts and 'All Accounts' for DisputesFilter, we can ensure that the value in the filter options and the value checked here is always the same.

glad-adyen: I think it's not the responsibility of this composable to guarantee the stability of the filter options. That remains the responsibility of the consumer (in this case), so that the options information/control is decoupled from the composable. The composable only guarantees state synchronization for the balance account filter.

However, it's possible that I don't yet understand the suggestion you are trying to pass across. Can you share a more concrete example, preferably highlighting the pros of your proposition?

packages/shared/composables-vue/src/useCursorPaginatedRecords.ts
```diff
@@ -0,0 +1,150 @@
...
+            void requestPage(cursor.value);
+        }
+    };
+
+    const goToPreviousPage = () => {
+        if (hasPrevious.value && previousCursor.value) {
+            page.value--;
+            void requestPage(previousCursor.value);
+        }
+    };
```
samuelsa-adyen, line 128: `goToNextPage` and `goToPreviousPage` both bump the page number before the request actually succeeds, and if the request fails, that update never gets rolled back. So if the page-two request fails, you end up in a weird state, page says you're on page two, but the records you're looking at are still from page one. And if you retry from there, it just increments the page again while re-requesting the same cursor.

packages/shared/hooks-preact/src/useTableColumns.ts
```diff
@@ -60,7 +60,7 @@ export const useTableColumns = <T extends string, C extends string>({
                 columnMap.set(current.key, {
                     ...existing,
                     ...current,
-                    visible: current.visibility !== 'hidden',
+                    visible: current.visibility === undefined ? existing.visible : current.visibility !== 'hidden',
```
samuelsa-adyen, line 63: Do we have this same logic in the vue implementation as well? just asking to keep parity

packages/shared/core/src/setupConfig.ts
```diff
@@ -43,10 +43,10 @@ export function createConfigController(
             let refreshing = session.context.refreshing;
             let unsubscribe = noop;
 
-            const updatePermission = (options?: { waitForSession?: boolean }) => {
+            const updatePermission = (options?: { waitForSession?: boolean; resetPermission?: boolean }) => {
```
samuelsa-adyen: nit: adding this new parameter to the options is causing that we're now passing `resetPermission` to `getPermission` and down the line. Could we maybe turn this into two different parameters? this could also simplify the passing down of `waitForSession`

packages/shared/hooks-preact/src/useTableColumns.ts
glad-adyen: Vue implementation follows a different approach towards achieving this, but the parity is now 1:1.

Some testing on [ReportsOverview](https://develop--platform-components.netlify.app/?path=/story/mocked-reports-reports-overview--data-customization&globals=viewport.value:mobile2) revealed this flaw in the Preact implementation. So when you pass a standard column in the fields list (without any column config overrides), the rendering of columns on mobile view (where we occasionally merge columns) is broken.

This change guarantees that custom columns are handled as expected on mobile (Preact).

packages/shared/composables-vue/src/useCursorPaginatedRecords.ts
glad-adyen: Nice catch!!!

I have modified the logic for adjacent page navigations to delegate to a new `navigate` function, which provides execution guarantees as well as page state rollback behavior.

packages/shared/core/src/setupConfig.ts
glad-adyen: Fair point. However, switching to two different `boolean` parameters in this case is (arguably) less clear from the consumer standpoint, and lacks self-documentation — e.g. what is the second parameter (since both parameters are boolean) without looking at the implementation?

Destructured `resetPermission` locally and forward only `getPermissionOptions`, while preserving the original call behavior when no permission options are set.

samuelsa-adyen (approved):

packages/shared/composables-vue/src/useBalanceAccountFilterState.ts
glad-adyen: Resolving this conversation for now. We can follow up as needed in subsequent revisions.

